PR #14, opened 2018-12-12
General SHC

AlexGabourie:
- Modified/added data structures, for eq. 3.114 in the GPUMD manual, to support any material. Previous code was optimized for graphene. 
- Changed 'int' to 'unsigned long long' to avoid overflow while indexing fv_all (based on previous bullet's changes).
- Updated conditions to allow for SHC + HNEMD


brucefan1983:
This is a very nice contribution. Thanks!

Issue #8, opened 2018-02-19
implementing a more general Tersoff potential


brucefan1983:
The current Tersoff potential in GPUMD is not of the most general form. I will make a more general version which can deal with most of the Tersoff-like potentials proposed in the literature. I also realized that there is still some room for improving the performance of the Tersoff and SW potentials. Therefore, there is a hope to have faster and more general Tersoff and SW potentials in GPUMD in the near future.

brucefan1983:
Now the Tersoff and SW potentials are more than 50% faster than before but the Tersoff potential is still not of the most general form. I will make the generalization some time later.

brucefan1983:
Now this has been done by Alexander J. Gabourie. Starting from this, it is easy to implement the version called tersoff/mod in LAMMPS. But that is another story. So I close this issue now.

PR #15, opened 2019-02-02
Tersoff1988

AlexGabourie:
Implements the Tersoff 1988 potential [J. Tersoff, Phys Rev B, 37, 6991 (1988).] in GPUMD (aimed to resolve issue #8 ). This is identical to 'pair_style tersoff' in LAMMPS (in output, not in implementation) and LAMMPS tersoff potential files can be easily converted to work with this new code. The potential also supports an arbitrary number of atom types This pull request also includes four new Tersoff potential parameter sets:

1. BN-C [A. Kinaci, et al.,  Phys Rev B, 86, 115410 (2012)]
2. Stanene [M. Cherukara et al, J Phys Chem Lett, 7, 3752 (2016)]
3. GaN [J. Nord, et al., J. Phys.: Condens. Matter, 15, 5649 (2003)]
4. SiC [J. Tersoff, Phys Rev B, 39, 5566 (1989) and errata (PRB 41, 3248)]

Attached is a PDF that contains the expression for the Tersoff 1988 potential, the potential file format, and the partial force expression (relevant to implementation in code). Additionally, there are folders for each of the potentials listed above. Each of these folders contains relevant files I used for testing. They also contain a figure that shows the absolute difference between GPUMD and LAMMPS forces (~11 decimal places is the output limit in f.out, so a difference 1e-11 is essentially equal). 

This code has only been tested to see if forces match LAMMPS and no substantial attempts have been taken to optimize it.

[tersoff1988.zip](https://github.com/brucefan1983/GPUMD/files/2823905/tersoff1988.zip)


brucefan1983:
Looks great!

PR #16, opened 2019-02-05
Tersoff1988 optimization

AlexGabourie:
This work is simply an optimization update to the Tersoff1988 potential. A switch to reading the Tersoff parameters from texture memory instead of global memory has been made as well as optimizations to the exponential term functions from the three-body interaction part of the potential. The potential still gives a force match to LAMMPS for four different, perturbed systems (SiC, BNC, GaN, Stanene; all used in original Tersoff1988 pull request https://github.com/brucefan1983/GPUMD/pull/15). 

A simple test script is used to gauge the performance benefit of the changes (excluding potential line):
velocity	        300
ensemble	nvt_ber 300 300 0.01
time_step	0.5
run 		10000

Performance: (system: old -> new -> % faster)
Stanene: 1.81763 Sec -> 1.23766 Sec -> ~47% faster
BNC: 2.47423 Sec -> 1.43466 Sec -> ~72% faster
SiC: 3.91685 Sec -> 1.76474 Sec -> ~122% faster
GaN: 30.0284 Sec -> 12.485 Sec -> ~140% faster

In addition to optimizations, some parameter checking has been added. This is a natural addition to the code for this pull request as we already need to ensure that the parameter _m_  is either 1 or 3 (might as well add the others).

### DIFF
--- a/src/tersoff1988.cu
+++ b/src/tersoff1988.cu
@@ -44,6 +44,8 @@ The version of the Tersoff potential as described in
     #define NINE_OVER_16  0.5625f
 #endif
 
+#define EPSILON 1.0e-15
+
 //Easy labels for indexing
 #define A 0
 #define B 1
@@ -77,6 +79,7 @@ Tersoff1988::Tersoff1988(FILE *fid, Atom* atom, int num_of_types)
     real *cpu_ters;
     MY_MALLOC(cpu_ters, real, n_entries*NUM_PARAMS);
 
+    char err[50] = "Error: Illegal Tersoff parameter.";
     rc = 0;
     int count;
 	double a, b,lambda, mu, beta, n, c, d, h, r1, r2, m, alpha, gamma;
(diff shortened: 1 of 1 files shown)


brucefan1983:
Alex, the speedups are really impressive. I have a few comments on some coding details:

1) `int m_int = int(m);`
Here using round() is safer. int() will truncate, which will cause problem if m = 2.999999999999...

2) `if (a < 0.0 || b < 0.0 || lambda < 0.0 || mu < 0 || beta < 0.0 || n < 0.0 ||
        		c < 0.0 || d < 0.0 || r1 < 0.0 || r2 < 0.0 || r2 < r1 ||
m_int - m != 0 || (m_int != 3 && m_int != 1) || gamma < 0.0)`

Here, I suggest change to checking the parameters one by one and reporting more exact error messages. I should also do it for the other potentials. After changing int() to round(), there is no need to check `m_int - m != 0`, and you only need to check whether int_m is 1, or 3.

3) `LDG(ters, i + ALPHA) == 0.0`
Here, using `==` to compare with zero is not safe. A safer way is to 
-- add `#define EPSILON 1.0e-15` in the beginning of this file
-- check whether `alpha < EPSILON` in the constructor
-- If yes, set `alpha = ZERO;` in the constructor
-- Then change the above expression to `LDG(ters, i + ALPHA) < EPSILON`

4) Similarly, you can change 
if (LDG(ters, i + M) == 3.0)
to
if (LDG(ters, i + M) > TWO)

AlexGabourie:
Thanks for the feedback. I've implemented your changes. The code still works and performance is similar. Otherwise it should be safer to run with better error messages.

brucefan1983:
Thanks Alex. I have just accepted this pull request. I am now implementing
triclinic box.

PR #28, opened 2019-03-13
Update VAC, DOS, SDC

AlexGabourie:
This pull request separates DOS (phonon Density of States) and SDC (Self Diffusion Coefficient) from each other and from the VAC (Velocity Autocorrelation) computation. The reason for this separation is because we want to have the [VAC be mass-weighted for the DOS](https://quantumwise.com/documents/manuals/ATK-2015/ReferenceManual/index.html/ref.velocityautocorrelation.html) and use only the raw VAC for SDC.  The new code has new DOS and SDC classes that deal specifically with their respective computations. The VAC code has removed the DOS and SDC code, but added additional checks for which type of VAC needs to be computed. Ultimately, the compute_vac command has been replaced by compute_dos and compute_sdc. The details of the new commands are outlined in the PDF included in the attached zip file.

The code has been tested on two systems graphene and MoS2. The graphene tests could be compared to ex2 included in the package. The tests were:
1. Graphene: Full system DOS -> matched ex2
2. Graphene: Use 'group' argument, one group for all atoms -> matched ex2
3. Graphene: Use 'group' argument, use group that is subset of total -> matched ex2
4. Graphene: Use 'num_dos_points' to output more DOS points (200 -> 300) -> matched ex2
5. Graphene: Use 'group' but using the second grouping method -> matched ex2

6. MoS2: Full system DOS -> consistent with literature
7. MoS2: 3 DOS runs in one run.in; first, Mo atoms only; second, S atoms only; third, all atoms -> consistent with literature
8. Graphene: SDC computation with group -> matched ex2

These tests are not exhaustive, but do test the common use cases I would expect for a user. The SDC command has been tested less than the DOS because most problems arose from the new VAC code, which DOS testing already captures. The SDC computation itself is relatively simple.

[update_vac.zip](https://github.com/brucefan1983/GPUMD/files/2963906/update_vac.zip)


AlexGabourie:
The PDF I attached incorrectly listed the contents of the sdc.out file. The attached PDF fixes this.
[gpumd_vac.pdf](https://github.com/brucefan1983/GPUMD/files/2968553/gpumd_vac.pdf)

brucefan1983:
Thanks for your great work! I am a little busy these few days. I will check
the code and document later. Perhaps we need to update the manual
accordingly?

I noticed that you used __ldg() for reading the velocity data. Does this
make it much faster?

Zheyong

AlexGabourie:
No problem, I enjoy working on this code. Yes, the manual will need to be updated. 

I did use the __ldg() function, but I didn't benchmark the speed. I believe there are other a couple of other spots where __ldg() can be used, but I didn't make those changes because I figured I should focus on getting the functionality correct. It should be easy to update.

brucefan1983:
It is usually a good habit to check the performance difference when
something like that is changed. For global memory access, if it is already
sequential, using __ldg()  usually does not bring about any speedup,
sometime even leading to the opposite.

AlexGabourie:
Ah, good point. While I won't get to it right away, I'll test the differences between using __lgd() and not and report back here.

brucefan1983:
I have integrated your document into the manual and updated ex2. It runs as expected. Thanks!

Issue #4, opened 2018-01-11
Add support for binary dump file


andeplane:
When simulating either large systems or long time scales, parsing a text file (`xyz.out`) may be very slow. I suggest at least supporting just a binary file that is easily loaded in numpy (Python) or similar.

brucefan1983:
OK. Good suggestion! I will try to do it. I have to figure out how to do it first :-)

andeplane:
What would be ideal is to actually use netcdf format (this is a binary format supported by many programs), but that would require that you link the code with their library. My guess is that you want to have a self contained code that just works out of the box. 

If you could show me how to add a command like
`dump_style text`, I could implement the binary output for you.

brucefan1983:
Yes, I don't want to make GPUMD depend on other packages, unless it is really unavoidable. For example, I have tried to use CUB (https://nvlabs.github.io/cub/) before, but switched to Thrust later. 

Making the output readable by other softwares like ovito is a very good idea. I did not do this because I did not need it in my previous research. I only used MATLAB to analysed my results :-)

It's very nice if you can contribute this part. There are many per-atom data which can be dumped, like velocities, forces, virials, and potentials. Perhaps the dump_style can apply to all these data?

To implement this, you need to 
1) add a variable such as "dump_style" to the Parameters structure in common.h so that you can use it later. 
2) add an else if to the parse() function in parse.cu and add a corresponding process_dump_style() function.
3) Modify (or rewrite) the code in the dump.cu file.

brucefan1983:
This has been solved by Alex.

PR #31, opened 2019-06-26
Output/netcdf

AlexGabourie:
I have added NetCDF output functionality (positions only) to partially address issue #4. This requires an external NetCDF library which violates one of the general goals of GPUMD, to be a standalone code. As such, this code will not be compiled unless the user enables the feature. All changes only affect the **dump_position** command although the original functionality is preserved. The revised command will have the following format:

**dump_position** *interval* <*optional_args*>

The *optional_args* provide additional functionality. Keywords can be in any order, but associated parameters must follow exactly. The new arguments are:
- *format* format
   format = {xyz, netcdf}
- *precision* <precision>
   precision = {single, double}
If *format* is not provided, the default value is 'xyz'. If *format* is 'xyz', the default output is the %g format in C. If precision is 'single', it uses the %0.9g format, and 'double' uses the %.17f format.
If *format* is 'netcdf', 'double' is default precision with double precision output. The 'single' format writes float32 values. The 'single' format does not follow official AMBER 1.0 conventions, but can still be read by many AMBER NetCDF readers.

So far, the GPUMD NetCDF output has been successfully read in VMD & OVITO and can also be read by ASE. It also stores the cell lengths and angles so that visualization software can illustrate boundaries of each cell as well as periodic copies.

For NetCDF, you need to build version [4.6.3](https://github.com/Unidata/netcdf-c/releases/tag/v4.6.3) configured with some extra flags. For example, my configure command looked like:

./configure --prefix=/home/alex/netcdf --disable-netcdf-4 --disable-dap

After the library is built, to use the NetCDF functionality with GPUMD, all you need to do is enable the USE_NETCDF flag (-DUSE_NETCDF on CFLAGS line in makefile) and provide the correct paths to the library. For example, my makefile had the following:

INC = -I/home/alex/netcdf/include
LDFLAGS = -L/home/alex/netcdf/lib
LIBS = -l:libnetcdf.a

The outputs of the XYZ and NetCDF have been (briefly) tested and show identical output positions. I also have run some simple tests to get an understanding on the time/size of outputs. The following table is constructed from different tests of the **dump_position** command for 3200 C atoms (graphene) run in NVT for 1000 steps. The position is output every single step. The results are:

| Test | Time (s) | File Size (MB) |
|-----------------------------|:-------:|:-------:|
| XYZ (GPUMD v2.4.1) | 3.16 | 86.2 |
| XYZ (default) | 3.17 | 86.2 |
| XYZ (single) | 3.94 | 115.0 |
| XYZ (double) | 6.42 | 210.8 |
| NetCDF (double) | 1.49 | 89.7 |
| NetCDF (single) | 1.39 | 51.3 | 

There are more details/nuances not listed here, but can be added here later or to the documentation.


brucefan1983:
Alex, this is a fantastic job.

PR #32, opened 2019-07-10
Support for an arbitrary number of potentials

AlexGabourie:
This work represents a large change to both the user input and the code. After some testing, I feel the code is in good enough shape to get some feedback. Hopefully you like it.

The new code allows the user to use as many potentials as they need to represent an arbitrarily complex system. This is compared to the current code that supports up to a few potentials. The new input removes the **potentials** keyword, adds the **potential_definition** keyword, and changes the **potential** keyword. It also removes the need for the 'layer' option and, as such, slightly changes the format of the xyz.in files. Finally, there is a modification on every single potential file definition, but it is very minor. The newly implemented commands are described below.

The **potential_definition** keyword is used to describe how you would like GPUMD to assign atoms to specific potentials. Now you can use both atom types and groups. It is the group feature that provides the flexibility to remove the 'layers' feature. There are two ways to use the new command:

**potential_definition** _type_ 
**potential_definition** _group_ _group_method_

_type_ and _group_  must be written verbatim and _group_method_ is an integer specifying which group method to use from the xyz.in file. This keyword must come before any **potential** keyword.

The new **potential** keyword is used to define ALL potentials (many-body and two-body). Each time the keyword is used, one potential is defined. The usage will look like:

**potential** _path_to_potential_ _[list of types or groups]_

Here the list will be integers denoting the types (or groups, if selected in the **potential_definition** line) that will participate in the potential. The types/groups must be listed in ascending order. For many-body potentials, the list must also be sequential. Types/groups for each potential must be in the same order and defined contiguously in the xyz.in file as well (i.e. a potential uses atom types 0 1, then 0 must be listed first in the xyz.in file with all atom type 1 definitions following, similar to current implementation). However, for two-body potentials, the list of numbers need not be sequential (i.e. 0 2 3 4 is acceptable). Finally, the order in which the **potential** keywords are used is not important. This will be illustrated in an example below.

With these new commands, there are few rules that are enforced:
1. An arbitrary number of potentials can be defined
2. Up to one many-body potential may be defined for each atom
3. An arbitrary number of two-body potentials may be defined for each atom
4. Each atom must participate in at least one potential

Rules 1 & 3 are not strictly true as there currently are hard-coded restrictions (10 potentials total), but that can be easily changed. (Also only up to 5 atom-type LJ potential definition is supported, this could also be changed).

To demonstrate the new usage, I will present a short script.

potential_definition type
potential /path/to/GPUMD/potentials/rebo/MoS_Liang_2009.txt 2 3
potential /path/to/GPUMD/potentials/tersoff/SiO_Munetoh_2007.txt 0 1
potential /path/to/GPUMD/test_num_pot/ljparams.txt 0 1 2 3
velocity 300
run 10000

As for the changes in each potential definition, I have changed: "tersoff_1989_1" to "tersoff_1989 1". Additionally, line 1 of the xyz.in file has changed from:
N M cutoff 0 has_velocity has_layer number_of_grouping_methods
to
N M cutoff 0 has_velocity number_of_grouping_methods

Since I had to make minor changes to every potential implementation, I have run a series of tests to confirm that the forces are identical. These runs also tested for ordering issues and grouping issues. So far, I have perfect agreement with the current code. The tests are:
- MoS2/SiO2 (with both LJ on and off) -> Tested rebo, tersoff 1989, and LJ
- SiO2/MoS2 (with both LJ on and off) -> Tested rebo, tersoff 1989, and LJ
- Cu (Same type, two groups, LJ off) -> Tested eam
- PbTe & Si -> Tested ri and sw
(materials separated by large gap. Didn't want materials to interact, simply wanted to test ordering issues in run.in and xyz.in. Also had to slightly modify current implementation slightly, but I believe it was a fair comparison)
- SiC & GaN -> Tested vashishta and tersoff 1988
(materials separated by large gap for the same reason as above)

Finally, I did a simple test to benchmark the overall speed of the new implementation. I ran 93,330 atoms for 10,000 steps (the code is the example above) for both the current implementation and the new code. I found that the new code runs ~<2% slower than the current implementation. Not ideal, but not too bad. This testing is by no means extensive. 

Please take a look at the new code, let me know if you have any question or concerns. I'm open to different implementation ideas or testing ideas.

PS: There was a minor error from the netCDF implementation. I fixed the final line in measure.cu.


brucefan1983:
This is a great improvement. A 2% performance drop is acceptable.

PR #33, opened 2019-07-13
Update phonon code for new commands & removed namespace command

AlexGabourie:
This is a minor update. The previous two pull requests (netCDF and arbitrary number of potentials) were only built and tested for the GPUMD code and did not consider the phonon code. This pull request addresses this issue. As of now, the phonon code can be built, but I have only tested it with the example in GPUMD/examples/phonon/ex1.  The test has given me identical results.

In addition to that, I have removed the 'namespace std' line and updated the relevant code that relied on the namespace definition.

Please let me know if I have missed anything or if you would like some additional testing.


brucefan1983:
Alex, thanks. Perhaps it's also good to update the inputs for the examples in examples/gpumd/, similar to the changes for the phonon example.

AlexGabourie:
Sure. That's a good idea. I'll try to update the examples very soon and add it to this request.

brucefan1983:
Alex, I have accepted this pull request. Could you make another pull request to update the examples in examples/gpumd/?

PR #40, opened 2019-09-14
Update random number generator seed

AlexGabourie:
For cluster GPUMD submissions, multiple simulations were starting during the exact same second causing the random number generators to output identical numbers. This small change now makes it such that the seed is based on nanoseconds instead of seconds, greatly reducing the chance of running identical simulations.


brucefan1983:
This is a good change. I actually used this strategy in my other GPU code, but forgot to do this for GPUMD.

PR #41, opened 2019-10-21
Green-Kubo Modal Analysis Implementation

AlexGabourie:
I have added a GPUMD implementation of the [Green-Kubo Modal Analysis (GKMA)](https://aip.scitation.org/doi/full/10.1063/1.5081722) method from Prof. Asegun Henry's group at MIT. The purpose of this implementation was to get the benefits of GPU acceleration, but also to use the formulation of heat flux used in GPUMD. While no formal benchmarks have been completed,  significant acceleration has been seen during informal testing due to the parallelizability of the GKMA method. 

The usage syntax is posted below but note that the commands are subject to change and the most up-to-date documentation will be posted in the [GPUMD documentation](https://gpumd.zheyongfan.org/index.php/Main_Page) in the future. The command will have the following format:

**compute_gkma** *sample_interval* *output_interval* *first_mode* *last_mode* *bin_option* *bin_size*

with argument definitions of:
- *sample_interval* = interval (in number of steps) used to compute the heat modal heat flux.
- *output_interval* = interval (in number of steps) to output modal heat flux. Heat flux is averaged over all samples per output interval.
- *first_mode* = First mode to include in calculation based on the eigenvector.out file.
- *last_mode* = Last mode to include in calculation based on the eigenvector.out file.
- *bin_option* = Which binning technique to use: 'bin_size' or 'f_bin_size' are current option. 
- *bin_size* = If *bin_option* = 'bin_size', then this is an integer describing how many modes are included per bin. If *bin_option* = 'f_bin_size', then binning by frequency is used and this is a float describing the bin size in THz.

Currently, this code also requires the use of an eigenvector file (eigenvector.out), similar to the original GKMA implementation (linked at top), that is generated by the phonon part of this GPU package. The modal heat flux is output to the 'heatmode.out' text file. For *n* bins, the output will look like (Ex: jx_outputNum_mode):

jx_1_1,j y_1_1, jz_1_1
jx_1_2, jy_1_2, jz_1_2
...
jx_1_n, jy_1_n, jz_1_n
jx_2_1, jy_2_1,j z_2_1
jx_2_2, jy_2_2, jz_2_2
...

So far, this code has been tested to give similar results to the LAMMPS implementation on small Si test systems. It also has been compared directly to and is in agreement with the full heat flux computed by GPUMD (i.e. compared GPUMD heat flux to the sum over all modal heat flux) to 1e-7. 

**Final notes and extras**:
- This code is very memory intensive and requires a lot of GPU memory and storage. Since the GPUMD code is currently only available for single-GPU systems, this may restrict the system size that is possible.
- There is a minor change in the hac.cu file that updates the EMD thermal conductivity calculation to match LAMMPS exactly. This does not invalidate any previously calculated results from the method.
- Once the method intputs/outputs are finalized, post-processing Python code will be available [here](https://github.com/AlexGabourie/thermo).
- Just to reiterate, this method and its arguments may be updated in the future based on the method's usage.

Please let me know if you have any questions, concerns, or changes you would like to make.

**Author(s)**
Alexander Gabourie (Stanford)


brucefan1983:
Very good work! So you also like unified memory! I Does it work on CC<6.0 (Kepler and Maxwell)? Sometimes a "cudaDeviceSynchronize()" is needed before accessing unified memory from the host (CPU) in CC<6.0. Forgetting this can result in "bus error".  No such problem for CC>=6.0 (Pascal, Volta, and Turing).

AlexGabourie:
Thank you! I figured I should try out unified memory if we eventually intend to work on a multi-GPU implementation. It was very easy to implement for the GKMA method. I only tested it for CC 7.5, but there is only one place where data moves from device to host (before writing to file) and I had to use the cudaDeviceSynchronize() command to get it working. Otherwise, after initialization, the rest of the work is kept on the GPU. Do you think that is sufficient, or should I test with different CC values?

brucefan1983:
If you have Kepler or Maxwell GPU, it's better to give it a test, because in these old cards, accessing unified memory from the host sometimes requires a cudaDeviceSynchronize() even if it seems to be unnecessary. This is one of the reasons why unified memory on new cards are more efficient.

AlexGabourie:
It doesn't look like I have access to Kepler or Maxwell GPUs... After checking, as far as I can tell, I can only get down to Pascal GPUs. If it really only is a problem when accessing unified memory from the host, then I think it should be OK because I already use the cudaDeviceSynchronize() function in the [only place that happens](https://github.com/brucefan1983/GPUMD/blob/dfbe2c8dd5caba2d7bef4c8352d55461606ecc3e/src/gkma.cu#L500).

brucefan1983:
Yes  then no problem!

AlexG <notifications@github.com> 于 2019年10月22日周二 08:09写道：

PR #42, opened 2019-10-23
GKMA - keyword syntax, heat current decomposition, output format specifier

AlexGabourie:
This is a simple update to the previously merged GKMA pull request (#41). The changes to the GKMA method are:

1. Removed the *output_interval* argument from the *compute_gkma* keyword. Averaging multiple heat flux samples per output seemed like a bad idea. Now there is one output for every sample at *sample_interval*.
2. Decomposed the heat flux into the in-plane and out-of-plane contributions for the *x* and *y* directions. This is the same decomposition used in the EMD and HNEMD methods.
3. Changed the output format specifier from %25.15e to %g. The GKMA heat flux only matched the true heat flux out to 1e-7 which should be captured with the %g specifier. This change can save space and I have not seen any noticeable changes in thermal conductivity calculations.

Below are some revisions of the usage and outputs:

**compute_gkma** *sample_interval* *first_mode* *last_mode* *bin_option* *bin_size*

with argument definitions of:
-   sample_interval = interval (in number of steps) used to compute the heat modal heat flux.
-   first_mode = First mode to include in calculation based on the eigenvector.out file.
-   last_mode = Last mode to include in calculation based on the eigenvector.out file.
-   bin_option = Which binning technique to use: 'bin_size' or 'f_bin_size' are current option.
-   bin_size = If bin_option = 'bin_size', then this is an integer describing how many modes are included per bin. If bin_option = 'f_bin_size', then binning by frequency is used and this is a float describing the bin size in THz.

For n bins, the output will look like (Ex: jx_outputNum_mode):
jxin_1_1, jxout_1_1, jyin_1_1, jyout_1_1, jz_1_1
jxin_1_2, jxout_1_2, jyin_1_2, jyout_1_2, jz_1_2
...
jxin_1_n, jxout_1_n, jyin_1_n, jyout_1_n, jz_1_n
jxin_2_1, jxout_2_1, jyin_2_1, jyout_2_1, jz_2_1
jxin_2_2, jxout_2_2, jyin_2_2, jyout_2_2, jz_2_2
...

Output will be found in heatmode.out.

Please let me know what you think of these changes or if you have any problems/suggestions.

### DIFF
--- a/src/gkma.cu
+++ b/src/gkma.cu
@@ -32,6 +32,7 @@ https://drive.google.com/open?id=1IHJ7x-bLZISX3I090dW_Y_y-Mqkn07zg
 #include <iostream>
 #include <sstream>
 
+#define NUM_OF_HEAT_COMPONENTS 5
 #define BLOCK_SIZE 128
 #define ACCUM_BLOCK 1024
 #define BIN_BLOCK 128
@@ -51,7 +52,7 @@ static __global__ void gpu_reset_data
     }
 }
 
-static __global__ void gpu_gkma_reduce
+static __global__ void gpu_gkma_reduce_xdotn
 (
         int num_participating, int num_modes,
         const real* __restrict__ data_n,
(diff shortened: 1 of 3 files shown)


brucefan1983:
I totally agree with these changes. 
1) Less options is better than more options.
2) The decomposition can be useful. This is a hard choice, but is also our special feature absent from other codes.
3) I think you can get an match up to 1.0e-15 if the eigenvectors.out from the phonon code used %25.15e instead of %g. Sorry for possible confusion :-) Anyway, %g is sufficient for this purpose.

PR #43, opened 2019-10-25
HNEMA - Modal Analysis for HNEMD

AlexGabourie:
This is an extension of the modal analysis technique to HNEMD. It uses many of the same techniques used in the GKMA code. 

The command will have the following format:

**compute_hnema** *sample_interval* *output_interval* *Fe_x* *Fe_y* *Fe_z* *first_mode* *last_mode* *bin_option* *bin_size*

with argument definitions of:
- *sample_interval* = interval (in number of steps) used to compute the heat modal heat flux.
- *output_interval* = interval (in number of steps) to output modal thermal conductivity. Thermal conductivity is averaged over all samples per output interval.
- *Fe_x*, *Fe_y*, *Fe_z* = x, y, z components of external driving force F_e in units of 1/Angstrom
- *first_mode* = First mode to include in calculation based on the eigenvector.out file.
- *last_mode* = Last mode to include in calculation based on the eigenvector.out file.
- *bin_option* = Which binning technique to use: 'bin_size' or 'f_bin_size' are current option.
- *bin_size* = If bin_option = 'bin_size', then this is an integer describing how many modes are included per bin. If bin_option = 'f_bin_size', then binning by frequency is used and this is a float describing the bin size in THz.

Currently, this code also requires the use of an eigenvector file (eigenvector.out) generated by this package. The modal thermal conductivity is output to the 'kappamode.out' file. For *n* bins, the output will look like (e.g. kx_output_num_mode):

kxi_1_1, kxo_1_1, kyi_1_1, kyo_1_1, kz_1_1
kxi_1_2, kxo_1_2, kyi_1_2, kyo_1_2, kz_1_2
...
kxi_1_n, kxo_1_n, kyi_1_n, kyo_1_n, kz_1_n
kxi_2_1, kxo_2_1, kyi_2_1, kyo_2_1, kz_2_1
...

This code has been confirmed to match the HNEMD output to 1e-2 if all modes are totaled. The lower matching accuracy is a symptom of lower accuracy output of the data in eigenvector.out. Agreement up to 1e-10 can be achieved if you change the output of the eigenvector.out file to %25.15e. 

**Author(s)**
Alexander Gabourie (Stanford)

### DIFF
--- a/src/hnema.cu
+++ b/src/hnema.cu
@@ -602,7 +602,7 @@ void HNEMA::process(int step, Atom *atom, Integrate *integrate, real fe)
     FILE *fid = fopen(hnema_file_position, "a");
     for (int i = 0; i < num_bins; i++)
     {
-        fprintf(fid, "%25.15f %25.15f %25.15f %25.15f %25.15f\n",
+        fprintf(fid, "%g %g %g %g %g\n",
                 bin_out[i], bin_out[i+num_bins], bin_out[i+2*num_bins],
                          bin_out[i+3*num_bins], bin_out[i+4*num_bins]);
     }
--- a/src/hnema.cu
+++ b/src/hnema.cu
@@ -1,0 +1,546 @@
+/*
+    Copyright 2017 Zheyong Fan, Ville Vierimaa, Mikko Ervasti, and Ari Harju
+    This file is part of GPUMD.
+    GPUMD is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    GPUMD is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


AlexGabourie:
One more note: This method can be used at the same time as the HNEMD method. Since both HNEMD and HNEMA methods define an external driving force, it is important to pay attention to which definition comes last as that will define the external driving force of the entire simulation.

brucefan1983:
Very good!  

A question: when you said "This code has been confirmed to match the HNEMD output to 1e-2 if all modes are totaled.", do you mean the **absolute** error is 1e-2. I guess the **relative** error is smaller?

AlexGabourie:
The error is the absolute error as in: Error = (HNEMD) - (sum of HNEMA modes). Since I was testing Si and the kappa values were large, I think that would mean that the relative error is smaller.

brucefan1983:
Yes, it's good to know. Thanks.


AlexG <notifications@github.com> 于 2019年10月26日周六 10:01写道：

PR #49, opened 2020-02-26
Combine GKMA & HNEMA into Modal Analysis + Single-precision implementation

AlexGabourie:
This combines the GKMA and HNEMA classes into a single MODAL_ANALYSIS class as much of their implementations used the same code. From a user's standpoint, nothing has changed. The commands listed in the [GKMA](42) and [HNEMA](#43) pull requests remain the same. Tests comparing the previous implementations and with this one show no difference in output or performance for both GKMA and HNEMA (double-precision calculation comparisons).

The MODAL_ANALYSIS class has also been updated to use single-precision only. Preliminary tests show modal analysis methods now run twice as fast with half the memory footprint; however, there is a decrease in accuracy. Initial tests have shown that the maximum absolute difference between double- and single-precision calculations is ~< 1e-4 and unlikely to noticeably affect any thermal conductivity calculations. I think the trade-off of accuracy for speed is justified in this case.

Please let me know if you have any comments, questions, or suggestions.

**Author(s)**
Alexander Gabourie (Stanford)


brucefan1983:
great work!

Issue #52, opened 2020-03-17
BUG: Error is raised when trying to compile the gpumd binary.


joabro:
BACKGROUND
============

It is currently not possible to compile the gpumd binary after cloning the GitHub repository.

WHAT HAPPENS
==============

The following error message is raised when trying to compile the gpumd binary:
```
force.cu(633): error: class "Measure" has no member "hnema"

1 error detected in the compilation of "/tmp/tmpxft_00002304_00000000-6_force.cpp1.ii".
make: *** [obj/force.o] Error 1
make: *** No rule to make target `obj_phonon/main_phonon.o', needed by `phonon'.  Stop.
```

joabro:
cc @brucefan1983 , @AlexGabourie

AlexGabourie:
Thanks for bringing this up. This is my fault, but the fix is easy. I will submit a pull request shortly.

AlexGabourie:
The pull request found [here](#53) should solve your problem.

joabro:
Unfortunately the [recent merge](https://github.com/brucefan1983/GPUMD/commit/6ff26614f15206a222616d6b028d071016540f20) not fix the problem and I still get the  same error message. The problematic line seem to be 633 in `force.cu`:
```
    // always correct the force when using the FCP potential
#ifdef USE_FCP
    if (!measure->hnemd.compute && !measure->hnema.compute)
    {
```

joabro:
Also the error appears when compiling `src/makefile` before I do `make  src/makefile.phonon`, in my install script:
```
#! /usr/bin/env bash
# JB, 12/14/2019

if [ "$#" -gt 0 ] ; then
    echo > /dev/stderr
    echo "usage: `basename $0` " > /dev/stderr
    echo "    Download and install GPUMD" > /dev/stderr
    echo "    using intel compilers, with netCDF" > /dev/stderr
    echo > /dev/stderr
    exit
fi

# Clone repository, if necessary
cwd=$PWD
currdir=`basename $cwd`
if [ "$currdir" != "GPUMD" ] ; then
    if [ ! -d "GPUMD" ] ; then
        git clone https://github.com/brucefan1983/GPUMD.git
    fi
    cd GPUMD
fi
cd src

# Load intel compilers
module load intelcuda/2019a

# Compile the gpumd binary
sed -i -e '/^CFLAGS/ s/$/ -DUSE_NETCDF -DUSE_FCP/' \
[...]

PR #53, opened 2020-03-18
Remove gkma and hnema. Add modal_analysis

AlexGabourie:
The phonon makefile was not updated in the [Modal Analysis pull request](#49), but has been here. This addresses issue #52. Thank you to @joabro for pointing out the mistake.


brucefan1983:
thanks!

PR #54, opened 2020-03-18
Update conditions for when USE_FCP defined

AlexGabourie:
Code (GPUMD and PHONON) now compiles with -DUSE_FCP flag. Fixes [modal_analysis bug](#52).


brucefan1983:
thanks for fixing this!

PR #55, opened 2020-05-03
Modal Analysis Optimization

AlexGabourie:
This work is limited to optimizing the [GKMA](#42) and [HNEMA](#43) methods. The changes are back-end and keyword usage is unchanged. I have swapped out the original custom modal analysis kernels and replaced them cuBLAS functions. The order of calculations has also been changed to reduce the number of read/write operations. Performance improvements vary depending on the size of systems, output rates, and number of bins, but initial testing has been promising. 

To give an example, an HNEMA calculation with the following conditions:
-  3680 atom graphene sheet (11040 modes)
-  output average thermal conductivity every 100 time steps 
-  52 bin output
- 10,000 step run

shows the following improvements:

|               |  Original  | Optimized | Reduction |
| --------  | --------- | ----------- | ---------- | 
| Memory | 1930 MB |   746 MB   |   ~62%    |
|    Time    |   69.4 s   |     21.6 s     |   ~69%   |  

with memory and time referring to that used during the HNEMA run in the simulation. Since the optimized version is more memory efficient, the difference between it and the original is expected to grow as the simulations scale in size.

Next, modal analysis now requires an *eigenvector.in* file as opposed to the *eigenvector.out* file used previously. There is no change to the file format, simply a change in the naming convention.

Please let me know what you think. Also, if you would like, I can try to update this code to use the GPU vectors. Otherwise, I will add those in a future pull request.

**Author(s)**
Alexander Gabourie (Stanford)

### DIFF
--- a/src/makefile
+++ b/src/makefile
@@ -11,7 +11,7 @@ CFLAGS = -std=c++11 -O3 -arch=sm_35 -DDEBUG
 
 INC =
 LDFLAGS = 
-LIBS = 
+LIBS = -lcublas
 
 SOURCES = main.cu gpumd.cu read_file.cu run.cu error.cu validate.cu \
 	mic.cu atom.cu velocity.cu neighbor.cu neighbor_ON1.cu neighbor_ON2.cu \
--- a/src/measure.cu
+++ b/src/measure.cu
@@ -817,11 +817,8 @@ void Measure::parse_compute_hnema(char **param, int num_param, Atom* atom)
         {
             PRINT_INPUT_ERROR("compute_hnema parameters must be positive integers.\n");
         }
-        int num_modes = h->last_mode - h->first_mode + 1;
-        if (num_modes % h->bin_size != 0)
-            PRINT_INPUT_ERROR("number of modes must be divisible by bin_size.\n");
         printf("    Bin by modes.\n"
-               "    bin_size is %d THz.\n", h->bin_size);
+               "    bin_size is %d modes.\n", h->bin_size);
     }
 
     // Hidden feature implementation
--- a/src/modal_analysis.cu
+++ b/src/modal_analysis.cu
@@ -86,90 +86,6 @@ static __global__ void gpu_scale_jm
 }
 
 
-static __global__ void gpu_reduce_xdotn
-(
-        int num_participating, int num_modes,
-        const float* __restrict__ data_n,
-        float* data
-)
-{
-    int tid = threadIdx.x;
-    int bid = blockIdx.x;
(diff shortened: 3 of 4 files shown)


brucefan1983:
This sounds very promsing. It seems now we have modified the same file during the same time and there are conflicts. Is it easy for you to resolve this?

AlexGabourie:
After merging the 'master' branch into MA_opt, my code no longer compiles. I will figure out the problem and fix it in the morning.

brucefan1983:
thanks. if you have any question, pleasse let me know. one change I made
yesterday is to replace LDG(a, n) to a[n]. I just know that LDG is not
needed when __restrict__ is present.

AlexG <notifications@github.com> 于 2020年5月3日周日 10:21写道：

AlexGabourie:
Many of the errors I'm getting are outside of my modal analysis changes. 
`error: identifier "FILE" is undefined`
is thrown for all code that uses FILE. Do you know how to fix this? Was this a problem for the current state of the master branch before I did a pull request?

brucefan1983:
I think it is because I cleaned up the common.cuh file. FILE is defined in
<stdio.h>. You can try to include this to the files that complains.

Zheyong


AlexG <notifications@github.com> 于 2020年5月3日周日 19:27写道：

brucefan1983:
It might be caused by my commits after your pull? If it is not a problem
for you, you can try to create a new pull request based on the latest
master?

Bruce Fan <brucenju@gmail.com> 于 2020年5月3日周日 19:44写道：

AlexGabourie:
Actually, all the commits from the master branch are in MA_opt already.  Just to double check, I downloaded a clean version of the master branch from the GPUMD GitHub page and I get the same errors. Can you get the master branch to compile?

brucefan1983:
I just tried and it compiles.

brucefan1983:
You mean you still get errors about FILE?

brucefan1983:
Perhaps I should test it in Linux. My recent developement was based on Windows.

brucefan1983:
Yes, I also got errors in Linux. 
So I will merge you pull request and fix the error.

AlexGabourie:
Ok, great! Thank you! 

I think I still have some LDG usage in my code but, when I update to using GPU vectors, I will remove them.

brucefan1983:
I have just added a few inclusion of <stdio.h> in a few files and it compiles in Linux now. 

You can go ahead and do the small changes you mentioned. If you have have question on the usages of GPU_Vector, just let me know.

brucefan1983:
After you finished, I would like to continue to re-factor the code.

PR #56, opened 2020-05-03
Modal Analysis with GPU_Vector

AlexGabourie:
This work updates the modal_analysis.cu and modal_analysis.cuh files to use the new GPU_Vector template class. No functionality changes are made and results are consistent with those from the previous [Modal Analysis Optimization](#55) pull request. 

Using the same 3680 atom graphene sheet from the [previous pull request](#55), I find that the code using GPU_Vectors runs ~1.6% slower than before. To me, this seems like a large penalty for a small change. Is this a reasonable difference in performance? If not, is it clear where I went wrong?

### DIFF
--- a/src/modal_analysis.cu
+++ b/src/modal_analysis.cu
@@ -72,7 +72,7 @@ static __device__ void gpu_bin_reduce
 (
        int num_modes, int bin_size, int shift, int num_bins,
        int tid, int bid, int number_of_patches,
-       const float* __restrict__ g_jm,
+       const float* __restrict__ jm,
        float* bin_out
 )
 {
@@ -92,11 +92,11 @@ static __device__ void gpu_bin_reduce
         int n = tid + patch * BIN_BLOCK;
         if (n < bin_size)
         {
-            s_data_xin[tid] += g_jm[n + shift];
-            s_data_xout[tid] += g_jm[n + shift + num_modes];
-            s_data_yin[tid] += g_jm[n + shift + 2*num_modes];
-            s_data_yout[tid] += g_jm[n + shift + 3*num_modes];
-            s_data_z[tid] += g_jm[n + shift + 4*num_modes];
+            s_data_xin[tid] += jm[n + shift];
+            s_data_xout[tid] += jm[n + shift + num_modes];
+            s_data_yin[tid] += jm[n + shift + 2*num_modes];
+            s_data_yout[tid] += jm[n + shift + 3*num_modes];
+            s_data_z[tid] += jm[n + shift + 4*num_modes];
         }
     }
 
@@ -130,7 +130,7 @@ static __global__ void gpu_bin_modes
        const int* __restrict__ bin_count,
        const int* __restrict__ bin_sum,
        int num_bins,
-       const float* __restrict__ g_jm,
+       const float* __restrict__ jm,
        float* bin_out
 )
 {
@@ -143,7 +143,7 @@ static __global__ void gpu_bin_modes
     gpu_bin_reduce
     (
            num_modes, bin_size, shift, num_bins,
-           tid, bid, number_of_patches, g_jm, bin_out
+           tid, bid, number_of_patches, jm, bin_out
     );
 
 }
@@ -155,7 +155,7 @@ static __global__ void elemwise_mass_scale
         const double* __restrict__ g_vx,
         const double* __restrict__ g_vy,
         const double* __restrict__ g_vz,
-        float* g_mv_x, float* g_mv_y, float* g_mv_z
+        float* mvx, float* mvy, float* mvz
 )
 {
     int i = blockIdx.x * blockDim.x + threadIdx.x;
@@ -164,12 +164,12 @@ static __global__ void elemwise_mass_scale
     {
         float sqrtmass = g_sqrtmass[i];
         float vx, vy, vz;
-        vx = __double2float_rn(LDG(g_vx, nglobal));
-        vy = __double2float_rn(LDG(g_vy, nglobal));
-        vz = __double2float_rn(LDG(g_vz, nglobal));
-        g_mv_x[i] = sqrtmass*vx;
-        g_mv_y[i] = sqrtmass*vy;
-        g_mv_z[i] = sqrtmass*vz;
+        vx = __double2float_rn(g_vx[nglobal]);
+        vy = __double2float_rn(g_vy[nglobal]);
+        vz = __double2float_rn(g_vz[nglobal]);
+        mvx[i] = sqrtmass*vx;
+        mvy[i] = sqrtmass*vy;
+        mvz[i] = sqrtmass*vz;
     }
 }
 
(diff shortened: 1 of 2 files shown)


brucefan1983:
Is this slow-down reproduciable? Have you tried it twice?

src/modal_analysis.cu
```diff
@@ -396,9 +448,9 @@ void MODAL_ANALYSIS::preprocess(char *input_dir, Atom *atom)
         }
 
         size_t eig_size = sizeof(float) * num_participating * num_modes;
-        CHECK(cudaMallocManaged((void **)&eig_x,eig_size));
-        CHECK(cudaMallocManaged((void **)&eig_y,eig_size));
-        CHECK(cudaMallocManaged((void **)&eig_z,eig_size));
+        eigx.resize(eig_size, Memory_Type::managed);
```
brucefan1983, line 451: Here, eig_size is the memory. Should use num_participating * num_modes instead. Otherwise more memory than needed is allocated. This can be a reason for slow-down.

src/modal_analysis.cu
```diff
@@ -416,8 +468,9 @@ void MODAL_ANALYSIS::preprocess(char *input_dir, Atom *atom)
         // Setup binning
         if (f_flag)
         {
-            double *f;
-            CHECK(cudaMallocManaged((void **)&f, sizeof(double)*num_modes));
+            GPU_Vector<double> f;
```
brucefan1983, line 471: These three lines can be written as 
GPU_Vector<double> f(num_modes, Memory_Type::managed);

src/modal_analysis.cu
```diff
@@ -479,57 +524,51 @@ void MODAL_ANALYSIS::preprocess(char *input_dir, Atom *atom)
...
+        size_t jmxyz_size = num_modes*3;
+        jmx.resize(jmxyz_size, Memory_Type::managed);
+        jmy.resize(jmxyz_size, Memory_Type::managed);
+        jmz.resize(jmxyz_size, Memory_Type::managed);
 
         num_heat_stored = num_modes*NUM_OF_HEAT_COMPONENTS;
-        size_t jm_size = sizeof(float) * num_heat_stored;
-        CHECK(cudaMallocManaged((void **)&jm,jm_size));
+        size_t jm_size = num_heat_stored;
+        jm.resize(jm_size, 0.0f, Memory_Type::managed);
```
brucefan1983, line 553: Here is another possible reason for slow-down. In the original version, the unified memory (managed memory) was first touched by GPU using the kernel function gpu_reset_data(). In the updated version, the unified memory was first touched by the CPU based on the function call resize(jm_size, 0.0f, Memory_Type::managed). The updated approach can lead to more page-fault conditions and cost some time. I will update the resize() function to allow it accepting a more optional argument that allow you choose where the data are initialized (CPU or GPU).

A final solution is to make choices for Windows and Linux code paths and add some optimization code for the Linux path. For example, the prefetch() function is very useful when using unified memory in Linux with CC>=6.0, but unfortunately, it is not supported in Windows for any GPU.

src/modal_analysis.cu
brucefan1983: Sorry, GPU_Vector<double> f(num_modes, Memory_Type::managed);

src/modal_analysis.cu
brucefan1983: Can you tried to switch back to the original kernel function to initialize the data and check if the speed is affected by the choice?

src/modal_analysis.cu
AlexGabourie: Good catch! That was a mistake. Updating that alone makes the code ~1% slower than #55.

src/modal_analysis.cu
AlexGabourie: Good idea. I'm going to simplify some of the resize calls too.

src/modal_analysis.cu
AlexGabourie: You are right that the new initialization will cause some page-faults. I switched back to the original kernel function, but it did not make a difference. Perhaps the array is too small to matter. 

At one point, I did try pre-fetching all of the GPU-relevant data in the preprocessing function. Unfortunately, the improvements were barely perceptible and the code was harder to read. I think this makes sense because most data is only loaded to the GPU once and then remains there until it is freed. Page-faults only occur for the first time the method is called.

For a sanity check, I revisited the code from #55 to double check the speeds. This time, the code runs slower than I reported in the pull request. It seems to be on par with the new code using GPU_Vector (after fixing the eig_size error you pointed out). After running 20+ tests, I find that there are some outlier simulations (fast and slow) that can mess with averages. At this point, I am satisfied and believe there is no real speed difference after updating to GPU_Vector.

AlexGabourie:
Thanks for the feedback, Zheyong!

brucefan1983:
Great. There should be no extra cost of  the GPU_Vector.

Zheyong

AlexG <notifications@github.com> 于 2020年5月4日周一 08:29写道：

brucefan1983:
If you want, you can try to make the handle of cublas as a class member of
modal_analysis and only create it once in preprocess and destroy it in
postprocess. I guess this may be faster.

Zheyong

Bruce Fan <brucenju@gmail.com> 于 2020年5月4日周一 08:48写道：

AlexGabourie:
Sure! I'll give it a try. Is the create/destroy process expensive? 

Since my code is nicely separated into independent x, y, and z-based calculations, I tried to use streams (one for each direction) but found my code to run a second or two slower. Now I wonder if creating and destroying the streams before and after the cublas functions was a problem. It's possible streams wouldn't help the example calculation in #55, but my GPU utilization (from nvidia-smi) was only ~86% and I wanted to see if I could push it to 100%.

brucefan1983:
You can try again. All these CUDA APIs are kind of expensive in
my experience.

Zheyong

AlexGabourie:
Wow! Again, fantastic suggestion. One run with the cublas handle as a class member reduced the run time to ~19 seconds (~2.6 seconds faster)!

It is a little late here, so I will try using streams again tomorrow morning. If I use the streams correctly, I should get equivalent performance to now in the worst case, right? For smaller systems, I should be able to squeeze out some extra performance.

brucefan1983:
I think so. Good night.

AlexG <notifications@github.com> 于 2020年5月4日周一 09:13写道：

AlexGabourie:
I tried using CUDA streams with the streams as class members (and initialized during preprocesssing). The performance took a hit. I tried the case where I used cublasSetStream_v2 to switch streams depending on the calculation and the case where I had multiple cublas handles that were fixed to different streams during preprocessing. Both cases has similar performance, but were worse than not using streams at all. Unless you have any suggestions, I don't think I will be adding streams at this time.

Just to make a note here, it should be possible to accelerate this method using the tensor cores. That may be an interesting addition in the future.

brucefan1983:
OK, good experiments!

AlexG <notifications@github.com> 于 2020年5月4日周一 19:42写道：

PR #57, opened 2020-06-04
combine host and device version of appy_mic

brucefan1983:
This PR combines the host and device version of the function which applies the minimum image convention.


AlexGabourie:
Good idea! That removes some repeated code. Since the mic function can be called from host or GPU, should it be called 'apply_mic' instead of 'dev_apply_mic'?

brucefan1983:
Good suggestion! I will change the function name.

PR #60, opened 2020-06-26
Improve codes related to defining multiple potentials

brucefan1983:
# Purpose
* The current version is not very efficient and convenient for some systems with inter-layer LJ interactions. 
* The aim of the pull request is to improve this part, hoping to produce a version that is more efficient and convenient. 

# Details
* Removed the `potential_definition` keyword. Potentials will always be defined in terms of atom types (the first column in `xyz.in`), although one can optionally modify the LJ potential using a grouping method (see below for details).
* The way of specifying non-LJ potentials is the same as in the master version.
* The LJ potential can be specified in one of the following two ways:
  * option A: `potential lj_potential_file`
  * option B: `potential lj_potential_file grouping_method_ID`
* In any case, the number of atom types in `lj_potential_file` should be the total number of types defined in `xyz.in`. 
* In option A, there is no modification to the LJ potential.
* In option B, the grouping method `grouping_method_ID` defined in `xyz.in` will be used to exclude LJ interactions between atoms with the same group label in grouping method `grouping_method_ID`.

# An example
* Consider a system consisting of 5 layers of materials. From bottom to top, they are MoS2, BN, BN, BN, and SiO2. That is, 3 layers of BN is embeded in one layer of MoS2 and one layer of SiO2.
* Using the master version, the potentials can be specified using the following commands in `run.in`:
```
potential_definition group 0
potential MoS_Liang_2009.txt 0 1
potential BN_Sevik_2011.txt 2 3
potential BN_Sevik_2011.txt 4 5
potential BN_Sevik_2011.txt 6 7
potential SiO_Munetoh_2007.txt 8 9
potential lj10.txt 0 1 2 3 4 5 6 7 8 9 
```
* Using the code from this PR, the potentials can be specified as follows:
```
potential MoS_Liang_2009.txt 0 1
potential BN_Sevik_2011.txt 2 3
potential SiO_Munetoh_2007.txt 4 5
potential lj6.txt 0 
```
* We see that the inputs for the the new version is simpler in this example. 
* I run 1000 steps with zero `time_step` (such that the positions are not changed) and output the forces for all the atoms:
```
ensemble    nve
time_step   0
compute     1 1000 1 force # grouping method 1 should be defined in xyz.in
run         1000
```

# Results
* The computed forces from the old and new versions match each other exactly. So the two versions are equivalent in term of computed data.
* The performance of the old version:
```
Number of neighbor list updates = 0.
Time used for this run = 19.233 s.
Speed of this run = 3.01607e+06 atom*step/second.
```
* The performance of the new version:
```
Number of neighbor list updates = 0.
Time used for this run = 15.229 s.
Speed of this run = 3.80905e+06 atom*step/second.
```
* So the new verison is about 27% faster than the old one in this example. 

# Summary
* The new version requres simpler user inputs, although in some cases there should be no difference.
* The new verison is usually faster, although in some cases there should be no difference in speed.


AlexGabourie:
This is a very nice change! It looks like it will result in a substantial speed improvement for many simulations. The new syntax is also clearly defined and straightforward. In your example, you tested the LJ potential with option B. Did you also try option A? I am curious how much of the speed improvement comes from the LJ neighborlist reduction and how much comes from the streamlined manybody potential evaluation. However, without seeing an option A test, I am confident that it would at least be correct. I think you should merge this with the main branch. Thank you for this update!

brucefan1983:
Option A cannot be used in this example (I will elaborate on this later). However, more details timing information could be obtained using `nvprof`. Here are the breakdown from my tests (laptop with RTX 2070; parts that have identical times are not shown):

Old time   New Time       Computation
6.1 s         3.9 s                finding local neighbor list
4.5 s         2.6 s                LJ force evaluation
6.6 s         6.4 s                Tersoff force evaluation

So you can see the most important part is actually the construction of the local neighbor list, because in the new version, it is only called 4 times in each step, while in the old version, it is called 6 times. The LJ force evaluation part is also affected a lot. The Tersoff part is actually not very much affected.

brucefan1983:
Here, option B is needed because the intralayer BB, BN, and NN interactions cannot be excluded using the types. We need to use a grouping method to achieve this. If there is only one BN layer between MoS2 and SiO2, using types only is sufficient. 

By looking at the timing details, I just realized that the local neighbor list construction is not efficiently done. Actually, we can use a single kernel to construct the local neighbor lists for each potential. This might result in another ~10% gain of performance, although the memory usage is increased. To achieve this, some refactoring of the code is needed I would like to try it out in a separate PR.

brucefan1983:
As for the Tersoff part, I think it also depends on the GPU model. My RTX 2070 does not have many CUDA cores and it can be easily saturated. Running with V100 can make a larger difference between the old and new verisons for the Tersoff part.

AlexGabourie:
Ah, right! Without changing the atom types, you must use option B for your example. Changing atom types for each BN layer to use option A would only recover the performance of the code on the current master branch. 

Your timing breakdown is very interesting. The Tersoff force evaluation is only for the hBN, right? (I think the SiO2 is also Tersoff, but the '89 implementation.) When I get GPU time, I can test differences between old and new on a V100.

As for the single-kernel neighbor list construction, do you think the memory usage increase is substantial? In my experience, GPUMD does not require a lot of memory (except when running HNEMA and GKMA calculations) compared to what is available on mid- to high-range GPUs. I would gladly trade performance for an increase in memory usage.

brucefan1983:
The Tersoff part is for both hBN and SiO2. Memory usage depends significantly on the cutoffs of the neighbor lists. A clever choice of the cutoffs usually requires inputs from the users. I found that many users who have used LAMMPS do not understand why `M` and `cutoff` are needed in `xyz.in` and have difficulty in understanding their meanings and in setting up them confidently. Neighor list is the most crucial part regarding memory usage and performance. This is why I let the users to set up these parameters in `xyz.in` and default to NOT update the neighor list. 

BTW, have you noticed that the initial forces in the example are very large? I think it is time to implement a minimizer. I would like to try sd and or/cg first. Of course, this is a separate work.

So I will first complete this PR. Thanks for your comments!

AlexGabourie:
There are approximately 4X as many SiO2 atoms as hBN atoms, so I would imagine that SiO2 contributes most of the time to the Tersoff force evaluation part. Your manybody evaluation speedup, which only affects the hBN in this example, will look better if we consider the time contribution of the SiO2 atoms. Based on the number of atoms in the system, we could roughly estimate the total SiO2 evaluation time to be ~5.1 s, meaning the hBN Tersoff evaluation is ~10-15% faster in this example with your hardware.

Even after using GPUMD for a while now, the `M` parameter is not very intuitive for me, so I tend to choose a larger M to be safe. For stationary solids, I don't need to worry about the neighborlist changing, which makes it easier for me. The `cutoff` I determine based on the potentials I include in my system. From a performance perspective, it makes sense to pre-allocate a maximum size neighborlist. The `cutoff` requirement only seems crucial for cases where the neighborlist is not updated. I may not understand the `cutoff` criteria fully. 

The structure I sent you came directly from the QuantumATK structure builder and was not relaxed at all yet. This particular structure took > 200,000 time steps for the initial pressure fluctuations to die down. A minimizer would be a fantastic feature to add.

PR #62, opened 2020-07-04
improve shc 

brucefan1983:
# Purposes
* Improve output related to the `compute_shc` command. Not only output SHC, but also output the Fourier transform of it. This will make post-process of the data simpler.


brucefan1983:
do not want to complete this PR now.

PR #66, opened 2020-07-13
Break dump_position into dump_position and dump_netcdf and add "group" option for the former

brucefan1983:
# Purpose
* Separate the class `DUMP_XYZ` and `DUMP_NETCDF` such that they are not dependent on each other. This is good for maintenance as `DUMP_NETCDF` is disabled by default.
* Add a `group` option to at least `DUMP_XYZ`.

# Implementation Details
* created a `Dump_Position` class and temporarily disabled `DUMP_NETCDF`. 
* The `group` option was added to `Dump_Position`.
* Added the `precision` option to `Dump_Position`.
* Simplified the code by moving some lines to some common parsing functions.
* `DUMP_XYZ` was removed. 
* Enabled `DUMP_NETCDF`.

# Change of inputs
* The `dump_position` keyword has been split into `dump_position` and `dump_netcdf`.
* The grammar of `dump_position` is
```
dump_position <interval> [group] <grouping_method> <group_id> [precision] <precision>
# Examples:
dump_position 10000
dump_position 10000 group 0 1 
dump_position 10000 precision double # default is %g in C
dump_position 10000 group 0 1 precision double
dump_position 10000 precision double group 0 1
```
* The grammar of `dump_netcdf` is
```
dump_netcdf <interval> [precision] <precision>
# Examples:
dump_position 10000
dump_position 10000 precision single # default is double
```


brucefan1983:
@AlexGabourie 
I have finished this PR. Could you try to compile the current code in this PR and test the `dump_netcdf` keyword?

AlexGabourie:
Sure! Let me do that now.

brucefan1983:
I just added a missing header. There might be other errors.

AlexGabourie:
There are a couple errors that I have found so far. First are some missing headers from dump_netcdf.cuh:
#include "model/box.cuh"
#include "utilities/read_file.cuh"

Also, there is a new term in the dump_netcdf files: `dump_interval_` which was previously `interval` and not all instances have been changed.
Finally, the `parse_precision` function is missing. I'm not sure where it is. Maybe you know what it's supposed to be now.

brucefan1983:
The error you mentioned should have been fixed. Could you try again?

AlexGabourie:
The netCDF code built with no errors and the output looks good from the couple tests I ran. 

The tests and results are here: [test_dump_netcdf.zip](https://github.com/brucefan1983/GPUMD/files/4920446/test_dump_netcdf.zip)
The way I view the results is to first load the POSCAR file, then the movie.nc file in VMD. Take a look and let me know if you want more tests.

brucefan1983:
I think the results should be ok as I only modified some interfaces, not
the internal things. If possible, you could test if the keyword behaves as
it should be for various valid/invalid parameters.

AlexG <notifications@github.com> 于 2020年7月14日周二 21:12写道：

AlexGabourie:
The keyword works well. It gives the proper error when incorrect and works as I expected when correct.

brucefan1983:
Thanks for your tests! I think it can be merged?

AlexG <notifications@github.com> 于 2020年7月14日周二 21:39写道：

AlexGabourie:
That sounds good to me! No issues to report from my end

PR #67, opened 2020-08-06
Calculate distance dependent DOS

brucefan1983:
# Purpose
Add code to calculate cross VAC (CVAC), which can be used to calculate the distance dependent DOS, which can then be used to calculate phonon coherence length.

# Reference
B. Latour, S. Volz, and Y. Chalopin, Microscopic description of thermal-phonon coherence: From coherent transport to diffuse interface scattering in superlattices, PHYSICAL REVIEW B 90, 014307 (2014).

# Usage
```
compute_cvac sample_interval num_correlation_steps <group> <group_id>
```

# Validation
I have confirmed that after summing up the contributions from the same particles, the VAC agree with that computed by using `compute_sdc`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 \.project
 core
 
+src/.vscode/*
 src/*/*.obj
 src/*/*.o
 src/gpumd*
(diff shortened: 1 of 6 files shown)


brucefan1983:
This calculation requires a lot of memory and also some nontrivial postprocessing. I think it is finished and thus merge it. Currently, only Haikuan is interested in using this feature.

PR #74, opened 2020-09-02
Add python scripts

brucefan1983:
# Purpose
I used MATLAB to preprocess and postprocess GPUMD. I want to learn to use Python now. In this PR, I will slowly add python scripts.


brucefan1983:
Alex will make a PR about this using his [thermo](https://github.com/AlexGabourie/thermo) package. So I think I can abandom this PR.

PR #76, opened 2020-12-09
Add a new potential in gpumd,but have some questions

Jonsnow-willow:
Add the files required for the lattice inversion potential.
Modify the code of the force.cu for lattice inversion potential.
Add the examples. 
The calculation of force and energy is accurate but the calculation of stress is wrong


brucefan1983:
Similar to the PR for GPUGA, Consider making smaller changes and only commit the necessary files. I will close this.

PR #82, opened 2021-01-29
Move some arrays to heap.

Lazemare:
Put an array with N float elements on stack, where N is the atom number of the simulated system, would trigger a stack overflow when N is very large. Thus I replaced the `float cpu_x[N]' statement with std::vector. This will fix the problem.

### DIFF
--- a/src/measure/dump_netcdf.cu
+++ b/src/measure/dump_netcdf.cu
@@ -331,9 +331,9 @@ void DUMP_NETCDF::write(
   if (precision == 1) // single precision
   {
     // must convert data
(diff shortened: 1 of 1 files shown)


brucefan1983:
Thanks for the contribution. @AlexGabourie , do you agree with this change?

AlexGabourie:
Yes, I think this is a good change. Thank you for updating this @Lazemare!

Lazemare:
> Yes, I think this is a good change. Thank you for updating this @Lazemare!

It's my pleasure! @AlexGabourie @brucefan1983

PR #83, opened 2021-02-11
Update DOS and SHC initialization on cpu

AlexGabourie:
# Issue
The "resize" function for std::vector only initializes elements to a value for new indices beyond previous size. For GPUMD simulations with multiple runs using the compute_shc and compute_dos keywords, the "allocate" function must be used to properly initialize some data structures.

## Note
The dos.cu code has been tested, but the shc.cu code has not. The change was simple enough that I did not feel it needed testing. Also, I believe that these are the only two files that need this change.


brucefan1983:
Alex, thanks for fixing thhis. I might have encountered problems caused by this, but I did not know the reason.

brucefan1983:
So my GPU_Vector behaves better than std::vector regarding data initilization :-)

PR #84, opened 2021-02-12
Binarize eigenvector.out/in

AlexGabourie:
# Purpose
The purpose of this pull request is to shrink the size of the eigenvector.out/in file. 

# Details
This file is generally very large (> 1GB), so its current text format is not useful for a user. The file is now updated to be in a binary, single-precision format. From my initial tests, a binary file's size is < 50% of a text file's size with more precision as well. The binary file is much faster to read (though read times are usually small compared to GKMA/HNEMA run times).

## Note
The format of the eigenvector.out/in file is identical [to before](https://gpumd.zheyongfan.org/index.php/The_eigenvector.in_input_file), but now flattened to a 1D array.


brucefan1983:
Alex, thanks for the improvement. Do you plan to update the user manual page:

https://gpumd.zheyongfan.org/index.php/The_eigenvector.in_input_file

AlexGabourie:
Yes, I will update the user manual now.

PR #87, opened 2021-05-19
2body 3body

brucefan1983:
I am experimenting a version that uses different cutoff radii for the radial and angular symmetry functions. After the superior performance is confirmed, I will merge it into master.


brucefan1983:
Using different cutoff radii for the radial and angular symmetry functions is very good for systems with long-range interactions, such as ionic crystals. Based on this, I merge this version into master.

PR #89, opened 2021-05-30
consider central atom type

brucefan1983:
This is an improvement over previous works.


brucefan1983:
This has been tested.

PR #90, opened 2021-06-29
Nep faster

brucefan1983:
Aim to make NEP a few times faster. Actaully the faster version aimed here has been prototyped in an earlier version, but it has been implemented up to l_max = 2 only. In this PR, I aim to extend it to l_max = 4 or 6.


brucefan1983:
I have made it up to l_max = 4, and achieved a speedup of more than 5x.  I think l_max = 4 is enough, if n_max is relatively large. So I decide to choose this faster version!

PR #123, opened 2021-12-25
Weight para

bigd4:
add weights of energy and virial in the nep.in

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 \.cproject
 \.project
 \.history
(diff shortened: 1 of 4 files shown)


brucefan1983:
Thanks for contribution. I happen to plan to change this part, using free-format nep.in. What you proposed here will be considered later. More importantly, the change you made involves a lot of changes to the examples/tutorials/manual. These should be done in a more comprehensive way. So I will close this PR without merging it.

Issue #146, opened 2022-02-03
Allow for choosing `l_max` in nep.in


brucefan1983:
Currently the only option is `l_max` = 4. Would be good to allow for choosing any number between 0 to 4. Useful for testing the importance of angular information.

brucefan1983:
This turns out to be not good:
1) there will be too many if statements that reduce the speed by about 10%.
2) `l_max` is really needed to converge the accuracyd. There is no point to try a smaller `l_max`. 

I cannot accept a 10% reduction of speed. So give up this idea.

PR #148, opened 2022-02-04
Let user to choose l_max

brucefan1983:


brucefan1983:
As explained in #146, this is not a good change I thus abandon it.

PR #150, opened 2022-02-10
allow for enabling/disabling 4-body

brucefan1983:
In `nep.in`, one will be able to enable/disable 4-body descriptors in the following way:
```
l_max 4 # only use 3-body
l_max 4 0 # only use 3-body
l_max 4 2 # use both 3-body and 4-body
```

### DIFF
--- a/src/force/nep3.cu
+++ b/src/force/nep3.cu
@@ -83,11 +83,13 @@ NEP3::NEP3(FILE* fid, char* input_dir, int num_types, bool enable_zbl, const Nei
   printf("    n_max_radial = %d.\n", paramb.n_max_radial);
   printf("    n_max_angular = %d.\n", paramb.n_max_angular);
 
-  count = fscanf(fid, "%s%d", name, &paramb.L_max);
-  PRINT_SCANF_ERROR(count, 2, "reading error for NEP potential.");
-  printf("    l_max = %d.\n", paramb.L_max);
+  int L_max_4body = 0;
+  count = fscanf(fid, "%s%d%d", name, &paramb.L_max, &L_max_4body);
+  PRINT_SCANF_ERROR(count, 3, "reading error for NEP potential.");
+  printf("    l_max_3body = %d.\n", paramb.L_max);
+  printf("    l_max_4body = %d.\n", L_max_4body);
 
-  paramb.num_L = paramb.L_max + 1;
+  paramb.num_L = L_max_4body == 2 ? paramb.L_max + 1 : paramb.L_max;
   paramb.dim_angular = (paramb.n_max_angular + 1) * paramb.num_L;
 
   int num_neurons2;
@@ -271,7 +273,11 @@ static __global__ void find_descriptor(
         }
         accumulate_s(d12, x12, y12, z12, gn12, s);
       }
-      find_q_nep3(paramb.n_max_angular + 1, n, s, q + (paramb.n_max_radial + 1));
+      if (paramb.num_L > paramb.L_max) {
+        find_q_with_4body(paramb.n_max_angular + 1, n, s, q + (paramb.n_max_radial + 1));
+      } else {
+        find_q(paramb.n_max_angular + 1, n, s, q + (paramb.n_max_radial + 1));
+      }
       for (int abc = 0; abc < NUM_OF_ABC; ++abc) {
         g_sum_fxyz[(n * NUM_OF_ABC + abc) * N + n1] = s[abc];
       }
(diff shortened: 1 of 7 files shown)


src/force/nep3.cu
```diff
@@ -83,11 +83,13 @@ NEP3::NEP3(FILE* fid, char* input_dir, int num_types, bool enable_zbl, const Nei
   printf("    n_max_radial = %d.\n", paramb.n_max_radial);
   printf("    n_max_angular = %d.\n", paramb.n_max_angular);
 
-  count = fscanf(fid, "%s%d", name, &paramb.L_max);
-  PRINT_SCANF_ERROR(count, 2, "reading error for NEP potential.");
-  printf("    l_max = %d.\n", paramb.L_max);
+  int L_max_4body = 0;
+  count = fscanf(fid, "%s%d%d", name, &paramb.L_max, &L_max_4body);
```
elindgren, line 87: Which file are these parameters read from? Is it the `nep.txt` file?

src/force/nep3.cu
```diff
@@ -271,7 +273,11 @@ static __global__ void find_descriptor(
         }
         accumulate_s(d12, x12, y12, z12, gn12, s);
       }
-      find_q_nep3(paramb.n_max_angular + 1, n, s, q + (paramb.n_max_radial + 1));
+      if (paramb.num_L > paramb.L_max) {
+        find_q_with_4body(paramb.n_max_angular + 1, n, s, q + (paramb.n_max_radial + 1));
+      } else {
+        find_q(paramb.n_max_angular + 1, n, s, q + (paramb.n_max_radial + 1));
```
elindgren, line 279: If I understand correctly, this function computes the descriptor q for a single atom, using pre-computed values for the basis functions( YLM & C4B)?

src/force/nep3.cu
```diff
@@ -457,7 +463,12 @@ static __global__ void find_partial_force_angular(
           gn12 += fn12[k] * annmb.c[c_index];
           gnp12 += fnp12[k] * annmb.c[c_index];
         }
-        accumulate_f12_nep3(n, paramb.n_max_angular + 1, d12, r12, gn12, gnp12, Fp, sum_fxyz, f12);
+        if (paramb.num_L > paramb.L_max) {
+          accumulate_f12_with_4body(
+            n, paramb.n_max_angular + 1, d12, r12, gn12, gnp12, Fp, sum_fxyz, f12);
+        } else {
+          accumulate_f12(n, paramb.n_max_angular + 1, d12, r12, gn12, gnp12, Fp, sum_fxyz, f12);
```
elindgren, line 470: And these are the forces on a single atom?

src/main_nep/nep3.cu
```diff
@@ -187,7 +187,11 @@ static __global__ void find_descriptors_angular(
         }
         accumulate_s(d12, x12, y12, z12, gn12, s);
       }
-      find_q_nep3(paramb.n_max_angular + 1, n, s, q);
+      if (paramb.num_L > paramb.L_max) {
+        find_q_with_4body(paramb.n_max_angular + 1, n, s, q);
+      } else {
+        find_q(paramb.n_max_angular + 1, n, s, q);
+      }
```
elindgren, line 194: What is the difference between `src/main_nep/nep3.cu` and `src/force/nep3.cu`? Does the former only predict energy, and the latter only forces?

elindgren (approved):
I think I can follow your changes somewhat, but I have a few questions.

src/force/nep3.cu
brucefan1983: Yes! YLM and C3B are \{C^{3b}_k\}_{k=0}^{23} and \{C^{4b}_k\}_{k=0}^{4} in the manuscript I sent you.

src/force/nep3.cu
brucefan1983: This is the "partial force"  dU_i/d \vec{r}_ij I defined in the manuscript  for atom i coming from atom j only. There will be summation over j in an outer loop (hence the name "accumulate")

brucefan1983: And it is only for the angular part, that is, the last two terns on the right hand side of Eq. (18) of the manuscript.

src/main_nep/nep3.cu
brucefan1983: src/main_nep/nep3.cu is used for the `nep` executable, that is, the NEP training code.
src/force/nep3.cu is used for the `gpumd` executable, that is, the MD production code.

They need to be different because they data structures are quite different (only a single box in MD but many structures in training).

src/main_nep/fitness.cu
```diff
@@ -192,7 +192,7 @@ void Fitness::report_error(
     }
     fprintf(fid_nep, "cutoff %g %g\n", para.rc_radial, para.rc_angular);
     fprintf(fid_nep, "n_max %d %d\n", para.n_max_radial, para.n_max_angular);
-    fprintf(fid_nep, "l_max %d\n", para.L_max);
+    fprintf(fid_nep, "l_max %d\n", para.L_max, para.L_max_4body);
```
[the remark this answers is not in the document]
elindgren: Ah I see!

src/force/nep3.cu
elindgren: Alright, makes sense :)

src/force/nep3.cu
brucefan1983: Actually it would be better to change the name YLM to C3B.

PR #159, opened 2022-02-24
Change to smooth weights A_ij = fc_ij instead of binary weights

elindgren:
Changes the graph weights A_ij = 1 to A_ij = fc(r_ij). For the PbTe system I haven't really seen an increase in performance, but I believe this will work better for more complex system (and it is more physically motivated). 

One could possibly normalize the weights fc(r_ij) such that, for instance, all rows in A sums to 1. With the current implementation, atoms with more weights will descriptors with larger magnitude, and may thus come to dominate.

This change costs a little bit of extra execution time. According to Ngsight, the execution time of the kernel `apply_gnn_message_passing` changes from ~415 µs to ~730µs, as well as a higher memory use. This translates to a total runtime of ~65s when running the PbTe example with the `nep.in` as shown below, compared to ~60s with A_ij = 1. 

Should be merged after #158

```bash
version      4
type         2 Te Pb
cutoff       6 6
n_max        12 6
neuron       40
batch        25
generation   1000
lambda_f     0
lambda_1     0.0002
lambda_2     0.0002
```

### DIFF
--- a/src/main_nep/nep4.cu
+++ b/src/main_nep/nep4.cu
@@ -249,6 +249,9 @@ static __global__ void apply_gnn_message_passing(
   const NEP4::ParaMB paramb,
   const NEP4::ANN annmb,
   const NEP4::GNN gnnmb,
+  const float* __restrict__ g_x12,
+  const float* __restrict__ g_y12,
+  const float* __restrict__ g_z12,
   const float* __restrict__ g_messages,
   const int* g_NN,
   const int* g_NL,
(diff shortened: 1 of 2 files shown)


src/main_nep/nep4.cu
```diff
@@ -261,18 +264,27 @@ static __global__ void apply_gnn_message_passing(
     // get messages for atom i and neighbors
     float q_theta_i[MAX_DIM] = {0.0f};
     float q_theta_j[MAX_NEIGHBORS * MAX_DIM] = {0.0f}; // maximum size when all atoms are neighbors
+    float fc_ij[MAX_DIM] = {0.0f};
```
brucefan1983, line 267: MAX_DIM should be MAX_NEIGHBORS

src/main_nep/nep4.cu
```diff
@@ -261,18 +264,27 @@ static __global__ void apply_gnn_message_passing(
...
+      int n2 = g_NL[index];
       for (int d = 0; d < annmb.dim; ++d) {
         q_theta_j[j + d * neighbor_number] = g_messages[n2 + d * N];
+        // Compute weight fc_ij
+        float r12[3] = {g_x12[index], g_y12[index], g_z12[index]};
+        // Can these be saved beforehand? Or is it too memory intensive?
+        float d12 = sqrt(r12[0] * r12[0] + r12[1] * r12[1] + r12[2] * r12[2]);
+        float fc12;
+        find_fc(paramb.rc_angular, paramb.rcinv_angular, d12, fc12);
+        fc_ij[j] = fc12;
```
brucefan1983, line 282: calculation of fc_ij can be outside of this loop over d, right after `int n2 = g_NL[index];`

src/main_nep/nep4.cu
elindgren: Good catch, I mixed them up in my head :rofl:

src/main_nep/nep4.cu
elindgren: Ah yes of course, now I'm basically doing it `N_des` times. I don't seem to have been very much awake this morning :laughing:

elindgren:
@brucefan1983 With your changes, this translates to a runtime of ~60s. I'd say this is ready to be merged now.

PR #160, opened 2022-03-02
Nep4 forces

elindgren:
Implementation of the forces for NEP4 as described in the overleaf document. 

Currently work in progress.

### DIFF
--- a/src/force/nep4.cu
+++ b/src/force/nep4.cu
@@ -86,33 +86,44 @@ NEP4::NEP4(FILE* fid, char* input_dir, int num_types, bool enable_zbl, const Nei
   printf("    l_max = %d.\n", paramb.L_max);
 
   int num_neurons2;
-  count = fscanf(fid, "%s%d%d", name, &annmb.num_neurons1, &num_neurons2);
+  count = fscanf(fid, "%s%d%d", name, &ann.num_neurons1, &num_neurons2);
   PRINT_SCANF_ERROR(count, 3, "reading error for NEP potential.");
 
-  rc = paramb.rc_angular; // largest cutoff
-
   paramb.rcinv_angular = 1.0f / paramb.rc_angular;
-  annmb.dim = (paramb.n_max_angular + 1) * paramb.L_max;
+  ann.dim = (paramb.n_max_angular + 1) * paramb.L_max;
 
-  printf("    ANN = %d-%d-1.\n", annmb.dim, annmb.num_neurons1);
+  printf("    ANN = %d-%d-1.\n", ann.dim, ann.num_neurons1);
 
-  annmb.num_para = (annmb.dim + 2) * annmb.num_neurons1 + 1;
-  printf("    number of neural network parameters = %d.\n", annmb.num_para);
+  ann.num_para = (ann.dim + 2) * ann.num_neurons1 + 1;
+  printf("    number of neural network parameters = %d.\n", ann.num_para);
   int num_para_descriptor =
     paramb.num_types * paramb.num_types * (paramb.n_max_angular + 1) * (paramb.basis_size + 1);
   printf("    number of descriptor parameters = %d.\n", num_para_descriptor);
-  annmb.num_para += num_para_descriptor;
-  printf("    total number of parameters = %d\n", annmb.num_para);
+  ann.num_para += num_para_descriptor;
+  gnn.num_para = ann.dim * ann.dim;
+  printf("    number of GNN parameters = %d.\n", gnn.num_para);
+  printf("    total number of parameters = %d\n", ann.num_para + gnn.num_para);
 
   paramb.num_types_sq = paramb.num_types * paramb.num_types;
-
-  nep_data.f12x.resize(neighbor.NN.size() * neighbor.MN);
-  nep_data.f12y.resize(neighbor.NN.size() * neighbor.MN);
-  nep_data.f12z.resize(neighbor.NN.size() * neighbor.MN);
-  nep_data.Fp.resize(neighbor.NN.size() * annmb.dim);
-  nep_data.sum_fxyz.resize(neighbor.NN.size() * (paramb.n_max_angular + 1) * NUM_OF_ABC);
-  nep_data.parameters.resize(annmb.num_para);
-
+  int N_times_max_neighbors = neighbor.NN.size() * neighbor.MN;
+  nep_data.f12x.resize(N_times_max_neighbors);
+  nep_data.f12y.resize(N_times_max_neighbors);
+  nep_data.f12z.resize(N_times_max_neighbors);
+  // nep_data.Fp.resize(neighbor.NN.size() * ann.dim);
+  // nep_data.sum_fxyz.resize(neighbor.NN.size() * (param.n_max_angular + 1) * NUM_OF_ABC);
+  // nep_data.parameters.resize(ann.num_para + gnn.num_para);
+  nep_data.dq_dx.resize(N_times_max_neighbors * ann.dim);
+  nep_data.dq_dy.resize(N_times_max_neighbors * ann.dim);
+  nep_data.dq_dz.resize(N_times_max_neighbors * ann.dim);
+  nep_data.q.resize(N2 * ann.dim);
+  nep_data.gnn_descriptors.resize(N2 * ann.dim);
+  nep_data.gnn_messages.resize(N2 * ann.dim);
+  nep_data.gnn_messages_p_x.resize(N2 * ann.dim * neighbor.MN);
+  nep_data.gnn_messages_p_y.resize(N2 * ann.dim * neighbor.MN);
+  nep_data.gnn_messages_p_z.resize(N2 * ann.dim * neighbor.MN);
+  nep_data.dU_dq.resize(N2 * ann.dim);
+  nep_data.s.resize(N2 * (paramb.n_max_angular + 1) * NUM_OF_ABC);
+  nep_data.parameters.resize(ann.num_para + gnn.num_para);
   update_potential(fid);
 }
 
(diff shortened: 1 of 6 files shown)


brucefan1983:
You don't need me first make dq/d r_ij available to you?

brucefan1983:
ok, then perhpas this is not ready for a PR. Or you might need to resolve conflicts later after my PR.

elindgren:
> ok, then perhpas this is not ready for a PR. Or you might need to resolve conflicts later after my PR.

Yes this is not ready to actually be a mergeable PR, but I wanted to have it as a PR so that we could take discussion etc. here. I will fix any conflicts after your PR :)

brucefan1983:
Ok, I see your points!

brucefan1983:
I will make the PR soon, and might also change something about memory usage . You followed my previous NEP2/NEP3 style to use a lot of `local memory` in CUDA kernel, but this might be not good for NEP4. Local memory would be good if it is used many times after its defintion; otherwise, it is better to use the global memory directly.

brucefan1983:
To clarify:

* `global memory` is also called device memory in CUDA, and is allocated in host (CPU) using `cudaMalloc()` or `cudaMallocManaged` in GPUMD.

* `local memory` refers to large arrays defined in CUDA kernel, such as `float q[MAX_DIM]`;

elindgren:
> I will make the PR soon, and might also change something about memory usage . You followed my previous NEP2/NEP3 style to use a lot of `local memory` in CUDA kernel, but this might be not good for NEP4. Local memory would be good if it is used many times after its defintion; otherwise, it is better to use the global memory directly.

Ah right, makes sense! I was actually getting a bit worried about memory usage, since I have been allocating quite a lot of arrays in the kernels.

brucefan1983:
> > I will make the PR soon, and might also change something about memory usage . You followed my previous NEP2/NEP3 style to use a lot of `local memory` in CUDA kernel, but this might be not good for NEP4. Local memory would be good if it is used many times after its defintion; otherwise, it is better to use the global memory directly.
> 
> Ah right, makes sense! I was actually getting a bit worried about memory usage, since I have been allocating quite a lot of arrays in the kernels.

Yes, memory usage is one of the most important things in CUDA programming. Different styles are needed for different algorithms.

elindgren:
@brucefan1983 I've made a rough first implementation of the more complicated forces for NEP4. I spent quite a lot of time trying to clean up the expressions, but I did not manage to come up with anything that helped us a lot. 

I hope you can follow my implementation roughly well. Following the notation in your derivations, this should be a direct implementation of \vec{F_i} = A+B = [messy expression]. See the pictures below.

Full expression F_i = A+B:

messy_full

Simplified expression (using `f_i`, `f_j` and `f_k` as in the code):

messy_simplified


There are two things that I'm unsure about. 1) I'm not sure that I'm, accessing dq_dr correctly in `apply_gnn_compute_messages` and 2) I'm unsure of how to include `find_force_angular` and `find_force_zbl` in the forces; from what I can tell, F_i should be computed according to our new derivations after `find_force_gnn`. Is this correct?

brucefan1983:
> There are two things that I'm unsure about. 1) I'm not sure that I'm, accessing dq_dr correctly in `apply_gnn_compute_messages` and 2) I'm unsure of how to include `find_force_angular` and `find_force_zbl` in the forces; from what I can tell, F_i should be computed according to our new derivations after `find_force_gnn`. Is this correct?

1) You have not accessed `dq_dx` etc correctly in `apply_gnn_compute_messages`, because these data involves neighbors. There is no neighbor information used.

2) As you have written `find_force_gnn`, the function `find_force_angular` can be removed. You can just leave the other function `find_force_zbl` there as it is a possible two-body contribution from ZBL.

brucefan1983:
You can also remove the definition and calling of `find_q_scaler` and `apply_q_scaler`. Currently you have scaled the descriptor but have not scaled the relevant derivatives. 

From my test of NEP3, it it not necessary to use this kind of descriptor normalization any more. The extra c-parameters will automatically select the various descriptor weights. Removing the normalization in NEP4 will make your life a little easier.

brucefan1983:
The force expression is really nontrivial. I also don't know whether we have derived it fully correctly. Let's see if it pass the finite-difference test (similar to the nep_cpu code).

elindgren:
> You can also remove the definition and calling of `find_q_scaler` and `apply_q_scaler`. Currently you have scaled the descriptor but have not scaled the relevant derivatives.
> 
> From my test of NEP3, it it not necessary to use this kind of descriptor normalization any more. The extra c-parameters will automatically select the various descriptor weights. Removing the normalization in NEP4 will make your life a little easier.

Alright, sounds good!

elindgren:
> > There are two things that I'm unsure about. 1) I'm not sure that I'm, accessing dq_dr correctly in `apply_gnn_compute_messages` and 2) I'm unsure of how to include `find_force_angular` and `find_force_zbl` in the forces; from what I can tell, F_i should be computed according to our new derivations after `find_force_gnn`. Is this correct?
> 
>     1. You have not accessed `dq_dx` etc correctly in `apply_gnn_compute_messages`, because these data involves neighbors. There is no neighbor information used.
> 
>     2. As you have written `find_force_gnn`, the function `find_force_angular` can be removed. You can just leave the other function `find_force_zbl` there as it is a possible two-body contribution from ZBL.

1. Right, makes sense! I will have another look at it.
2. Perfect, thanks for the clarification!
3.

elindgren:
@brucefan1983 I fixed NEP4 to use the neighbor information in dq_dr. Next step would be to check the forces using finite differences. I checked your code in `validate_force` in the `gpumd` module, but I didn't find it entirely obvious how to implement something similar in the `nep` module. How do you suggest I should go about it? :)

brucefan1983:
After you implememt NEP4 in the GPUMD part (`force/nep4.cu` and `force/nep4.cuh` and `force/nep4_small_box.cuh`), you only need to compile the code by adding `-DVALIDATE_FORCE` to makefile. This will enable the following code in main_gpumd/run.cu:

```
// only for my test
#ifdef VALIDATE_FORCE
  validate_force(
    box, atom.position_per_atom, group, atom.type, atom.potential_per_atom, atom.force_per_atom,
    atom.virial_per_atom, neighbor, &force);
#endif
```

Then using a `run.in` like the following, 
```
potential   examples/nep_potentials/PbTe/train/nep.txt 0 1
velocity    1

ensemble nve
time_step 0
dump_force 1
run 1
```

you will get a file named `force_compare.out` in the main folder of GPUMD.  The first half rows will be analytical forces and the second half are finite-difference forces. Using single precision, you can expect they match up to 1.0e-3 if there is no error in force implementation.

brucefan1983:
> 

I warned you about the speed :-)

brucefan1983:
So you need to implement NEP4 in the GPUMD part too. You need to pay attention to some subtle differences between the GPUMD and the NEP parts!

elindgren:
> So you need to implement NEP4 in the GPUMD part too. You need to pay attention to some subtle differences between the GPUMD and the NEP parts!

Alright, I'll get started with that then!

> I warned you about the speed :-)

I know, but I didn't think it would be this bad :sweat_smile:

elindgren:
@brucefan1983 I have begun working on implementing NEP4 in `force/nep4`. I have a few questions.

1. I don't want to copy the code for NEP4 from `main_nep` to `force`, since that would make it much harder to maintain. Rather, I've declared the NEP4 in  `force` (now called `NEP4MD` instead of `NEP4`) as inheriting NEP4 from `main_nep`. However, I'm running into the issue that `main_nep` uses single precision floats, while `force` uses doubles. Do you know if it is possible to overload the functions from `main_nep` so they use doubles, without having to reimplement them (basically just changing the types in the function call)? Or is there a reason as to why I would need to reimplement the functions in `force`?

2. What are the `N1` and `N2` parameters? Do they have to do with domain decomposition?

3. When allocating the GPU vectors in `main_nep` I have used `N` for the number of atoms. However, during MD I assume `N` can be millions of atoms. Should I use something like `N1` or `N2` instead when allocating memory?
```c++
  nep_data.dq_dx.resize(N_times_max_NN_angular * ann.dim);
  nep_data.dq_dy.resize(N_times_max_NN_angular * ann.dim);
  nep_data.dq_dz.resize(N_times_max_NN_angular * ann.dim);
  nep_data.q.resize(N1 * ann.dim);
  nep_data.gnn_descriptors.resize(N1 * ann.dim);
  nep_data.gnn_messages.resize(N1 * ann.dim);
  nep_data.gnn_messages_p_x.resize(N1 * ann.dim * MAX_NEIGHBORS);
  nep_data.gnn_messages_p_y.resize(N1 * ann.dim * MAX_NEIGHBORS);
  nep_data.gnn_messages_p_z.resize(N1 * ann.dim * MAX_NEIGHBORS);
  nep_data.dU_dq.resize(N1 * ann.dim);
  nep_data.s.resize(N1 * (param.n_max_angular + 1) * NUM_OF_ABC);
  nep_data.parameters.resize(ann.num_para + gnn.num_para);
```

brucefan1983:
I don't want to copy the code for NEP4 from main_nep to force, since that would make it much harder to maintain. Rather, I've declared the NEP4 in force (now called NEP4MD instead of NEP4) as inheriting NEP4 from main_nep. However, I'm running into the issue that main_nep uses single precision floats, while force uses doubles. Do you know if it is possible to overload the functions from main_nep so they use doubles, without having to reimplement them (basically just changing the types in the function call)? Or is there a reason as to why I would need to reimplement the functions in force?

The implementation of NEP in main_nep/ and force/ are not identical. When you get into the details, you will notice some differences. So we need to duplicate some code. But later when you are sure everything is fine, you can put as much commond code as possible in, say, nep_utilities.cuh.

brucefan1983:
What are the N1 and N2 parameters? Do they have to do with domain decomposition?

They are related empirical potentials and I have not changed the interface for NEP. For mixed empirical potentials such as Tersoff + SW + ..., one might want to apply Tersoff to atoms [N1, N2), and SW to atoms [N2, N3), ...

So N1 and N2 are the boundary indices for the current potential. For NEP, N1 will be 0 and N2 will be N (total number of atoms).

brucefan1983:
When allocating the GPU vectors in main_nep I have used N for the number of atoms. However, during MD I assume N can be millions of atoms. Should I use something like N1 or N2 instead when allocating memory?

As explained above, in NEP, N1=0 and N2=N, and yes, we need to allocate memory for N atoms. At least NEP2/NEP3 can deal with several millions of atoms using A100 GPU. It is your task to keep the memory usage in NEP4 under a reasonable limit.

brucefan1983:
I think we need to first focus on the **correctness** (energy and force much be consistent) and **accuracy** (we need to see an improved accuracy compared to NEP3 for at least some systems), then **speed** and **memory** usage, and last **maintainability**.

brucefan1983:
Also note that there is normal path and a "small box" path in force/ for any NEP. This means triple duplication!

The code for "small box" is similar to that in main_nep, which used atomicAdd(), while that in the normal path has not used atomicAdd(). This is a very subtle difference you need to pay attention to.

elindgren:
> The implementations of NEP in main_nep/ and force/ are not identical. When you get into the details, you will notice some differences. So we need to duplicate some code. But later when you are sure everything is fine, you can put as much common code as possible in, say, nep_utilities.cuh.

Alright, makes sense!

> I think we need to first focus on the correctness (energy and force much be consistent) and accuracy (we need to see an improved accuracy compared to NEP3 for at least some systems), then speed and memory usage, and last maintainability.

Yes I agree :+1: 

> Also note that there is normal path and a "small box" path in force/ for any NEP. This means triple duplication!
> 
> The code for "small box" is similar to that in main_nep, which used atomicAdd(), while that in the normal path has not used atomicAdd(). This is a very subtle difference you need to pay attention to.

I had not noticed the difference regarding `atomicAdd()`, thanks for the heads-up!

elindgren:
To get this to work, we need to reevaluate the approach. Let's close this for now.

PR #161, opened 2022-03-03
prepare for NEP4 force

brucefan1983:
* speed up the kernel `apply_gnn_message_passing`
* speed up the kernel `apply_gnn_compute_message`
* make { dq^i_mu/d\vec{r}_ij } available. This array is of length `3 * num_atoms * MAX_NUM_NEIGHBORS * dim`. In NEP4, `cutoff` can be small (as the effective cutoff will be doubled by message passing), and `MAX_NUM_NEIGHBORS ` can be small, such as `~20`. In NEP4, `dim` can also be small. Anyway, NEP4 will be more time-consuming and memory consuming than NEP2/NEP3, but hope it is more accurate.

`nep.in` for the PbTe example:
```
version      4
type         2 Te Pb
cutoff       6 6 # only the second number (angular cutoff) is used for NEP4
n_max        6 6 # only the second number (angular n_max) is used for NEP4
neuron       40
batch        25
generation   100
lambda_f     0
lambda_1     0.001
lambda_2     0.001
```

### DIFF
--- a/src/main_nep/nep4.cu
+++ b/src/main_nep/nep4.cu
@@ -97,7 +97,7 @@ static __global__ void find_descriptors_angular(
   const float* __restrict__ g_y12,
   const float* __restrict__ g_z12,
   float* g_descriptors,
-  float* g_sum_fxyz)
+  float* g_s)
 {
   int n1 = threadIdx.x + blockIdx.x * blockDim.x;
   if (n1 < N) {
@@ -129,7 +129,7 @@ static __global__ void find_descriptors_angular(
       }
       find_q(paramb.n_max_angular + 1, n, s, q);
       for (int abc = 0; abc < NUM_OF_ABC; ++abc) {
-        g_sum_fxyz[(n * NUM_OF_ABC + abc) * N + n1] = s[abc];
+        g_s[(n * NUM_OF_ABC + abc) * N + n1] = s[abc];
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/main_nep/nep4.cu
```diff
@@ -229,122 +232,88 @@ static void __global__ find_max_min(const int N, const float* g_q, float* g_q_sc
...
-  const NEP4::GNN gnnmb,
+  const NEP4::Para para,
+  const NEP4::ANN ann,
   const float* __restrict__ g_x12,
   const float* __restrict__ g_y12,
   const float* __restrict__ g_z12,
   const float* __restrict__ g_messages,
   const int* g_NN,
   const int* g_NL,
-  float* __restrict__ gnn_descriptors)
```
elindgren, line 258: Why do you remove the `__restrict__` keyword here?

src/main_nep/nep4.cu
```diff
@@ -229,122 +232,88 @@ static void __global__ find_max_min(const int N, const float* g_q, float* g_q_sc
...
+      float q_out_nu = g_messages[n1 + nu * N]; // fc(r_ii) = 1
+      // TODO perhaps normalize weights? Compare Kipf, Welling et al. (2016)
+      for (int j = 0; j < num_neighbors_of_n1; ++j) {
+        int index = n1 + N * j;
+        int n2 = g_NL[index];
+        float r12[3] = {g_x12[index], g_y12[index], g_z12[index]};
+        float d12 = sqrt(r12[0] * r12[0] + r12[1] * r12[1] + r12[2] * r12[2]);
+        float fc12;
+        find_fc(para.rc_angular, para.rcinv_angular, d12, fc12);
+        q_out_nu += fc12 * g_messages[n2 + nu * N];
```
elindgren, line 273: So you move the kernels I wrote from `nep_utilities.cuh` to `nep4.cu`?

elindgren (approved):
Looks good :+1:  I just had two questions/comments.

src/main_nep/nep4.cu
brucefan1983: Yes, I moved two `__device__` functions in `nep_utilities.cuh` to `main_nep/nep4.cu`. 

The `__device__` functions in `nep_utilities.cuh` are supposed to be called by functions in both `main_nep/nepX.cu` and `force/nepX.cu` (`X` = 3 or 4). 

However, to make the CUDA kernels in `main_nep/nep4.cu` more cunderstandable (all the relevant code in the same function), I think it is better to write all the code in the kernels first and condier re-factoring in the future.

Issue #169, opened 2022-03-29
Do we need a BAOAB-splitting Langevin integrator?


Lazemare:
The current Langevin integrator implemented in GPUMD uses the OBABO-splitting reported by Bussi _et al._ in [1]. But according to Leimkuhler _et al._ [2], the Langevin integrator with a BAOAB-splitting could provide much better stability and configuration-sampling ability when using a large time step. So this splitting methodology could be useful in some cases, especially when the users are interested in the conformational changes of complex systems. If we need this new integrator, I would like to open a PR and work around it.

```
[1] Bussi, G., & Parrinello, M. (2007). Accurate sampling using Langevin dynamics. Physical Review E, 75(5), 056707.
[2] Fass, J., Sivak, D. A., Crooks, G. E., Beauchamp, K. A., Leimkuhler, B., & Chodera, J. D. (2018). Quantifying configuration-sampling error in Langevin simulations of complex molecular systems. Entropy, 20(5), 318.
```

brucefan1983:
Good suggest, go ahead if you are interested in this.

Lazemare:
Thanks a lot! I have written the code to implement the new integrator, you can find them in my GPUMD fork. The new integrator is named `ENSMBLE_BAO` and could be invoked from the input file with the keyword `nvt_bao`. Due to the different splitting scheme, I could not use the `velocity_verlet` function from the base class. So I divided this function into different parts and put them in the new source file. I think this is something that could be improved.

And here are my test results: I tested the new integrator on a 20×20×20 silicon lattice system (with 428 atoms in it), under a temperature of 3500 K. I run multiple simulations using different timesteps. After these simulations, I calculated the RDF curves from the trajectories. There is the `run.in` file:
```
potential     ../../../potentials/tersoff/Si_Fan_2019.txt 0
velocity      300

ensemble      npt_ber 300  3500 100 0 0 0 53.4059 53.4059 53.4059 2000
time_step     1
dump_thermo   100
run           200000

ensemble      npt_ber 3500 3500 100 0 0 0 53.4059 53.4059 53.4059 2000
time_step     1
dump_thermo   100
run           50000

ensemble      nvt_bao 3500 3500 400
time_step     0.5
dump_thermo   100
dump_position 200
run           1000000
```
First, I checked that the new integrator could generate the same RDF curves as the RDF curves generated by the OBABO integrator implemented in GPUMD (I know they look poor since the simulation time is short) :
[...]

PR #173, opened 2022-04-05
Implement the BAOAB Langevin integrator.

Lazemare:
Implement the BAOAB Langevin integrator we discussed in https://github.com/brucefan1983/GPUMD/issues/169.

### DIFF
--- a/src/integrate/ensemble_bao.cu
+++ b/src/integrate/ensemble_bao.cu
@@ -1,0 +1,542 @@
+/*
+    Copyright 2017 Zheyong Fan, Ville Vierimaa, Mikko Ervasti, and Ari Harju
+    This file is part of GPUMD.
+    GPUMD is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    GPUMD is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+    You should have received a copy of the GNU General Public License
+    along with GPUMD.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+/*----------------------------------------------------------------------------80
+The Langevin thermostat with the BAOAB splitting:
+[1] Leimkuhler, Benedict, and Charles Matthews. Applied Mathematics Research
+    eXpress 2013.1 (2013): 34-56.
+[2] Fass, Josh, et al. Entropy 20.5 (2018): 318.
+------------------------------------------------------------------------------*/
+
+#include "ensemble_bao.cuh"
+#include "utilities/common.cuh"
+#define CURAND_NORMAL(a) curand_normal_double(a)
+
+// initialize curand states
+static __global__ void initialize_curand_states(curandState* state, int N)
+{
+  int n = blockIdx.x * blockDim.x + threadIdx.x;
+  // We can use a fixed seed here.
+  if (n < N) {
+    curand_init(123456, n, 0, &state[n]);
+  }
+}
+
+Ensemble_BAO::Ensemble_BAO(int t, int fg, int N, double T, double Tc)
+{
+  type = t;
+  fixed_group = fg;
+  temperature = T;
+  temperature_coupling = Tc;
+  c1 = exp(-1.0 / temperature_coupling);
+  c2 = sqrt((1 - c1 * c1) * K_B * T);
+  curand_states.resize(N);
+  int grid_size = (N - 1) / 128 + 1;
+  initialize_curand_states<<<grid_size, 128>>>(curand_states.data(), N);
+  CUDA_CHECK_KERNEL
+}
+
+Ensemble_BAO::Ensemble_BAO(
+  int t,
+  int fg,
+  int source_input,
+  int sink_input,
+  int source_size,
+  int sink_size,
+  int source_offset,
+  int sink_offset,
+  double T,
+  double Tc,
+  double dT)
+{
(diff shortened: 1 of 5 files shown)


brucefan1983:
A general comment:

There are many idential CUDA kernels used in both `ensemble_bao.cu` and `ensemble_lan.cu`. You can consider moving these common kernels to a file `langevin_unitilities.cuh". Then just include this file to both `ensemble_bao.cu` and `ensemble_lan.cu`. See an example for the file `npt_utilities.cuh`, which is included by both `ensemble_npt_ber.cu` and `ensemble_npt_src.cu`. 

Before you do this refactoring, compile the code with `-DDEBUG` and run your example and save the outputs. After you do this refactoring, run the example again and you should get exactly the same outputs.

Lazemare:
Quite reasonable! Let me change the code and make more tests.

Lazemare:
I moved these identical kernels functions to the new `langevin_utilities.cuh` header file. The code is much cleaner now.

brucefan1983:
Great. You can fix my other review comments and particularly the one related to "fixed group".

Lazemare:
I'm sorry, but it seems that I can not find your other comments anywhere. Did I miss something?

brucefan1983:
They are specific comments about some lines of code. Like this:

image

brucefan1983:
In some sense, these comments are "pending", and you are supposed to "close" them with appropriate actions.

Lazemare:
That's weird, your reviews are not showing on my PR page
Screenshot from 2022-04-08 17-32-27

brucefan1983:
you can check "files changed"

Lazemare:
Yeah, I have checked that page, but they are not there too
Screenshot from 2022-04-08 17-42-40

brucefan1983:
Perhaps I need to add you as a collaborator. I just did it. Could you check
again?

Lazemare:
I checked these pages again, but still, I can not find the comments 😢

brucefan1983:
OK, never mind, you can just check this question: 

For operator_B, you considered the case with fixed group, but why not consider it for operator_A as well?

Could this be the reason for the bad temperature control in your NEMD test?

Lazemare:
OK, let me explain the `operator_A` function and the `fixed_group` variable.
The names `operator_A` and `operator_B` come from formula 5 in [1]. The `operator_A` function is corresponding to an operator that advances the positions:

<p align="center">
    
</p>

While the `operator_B` function is corresponding to an operator that advances the velocities:

<p align="center">
    
</p>

(The `O` operator is corresponding to an Ornstein–Uhlenbeck process.)
Using such notations, the standard Velocity-Velet integrator reads as:

<p align="center">
    
</p>

And the BAOAB Langevin integrator reads as:

<p align="center">
    
</p>

To implement these operators, I looked into the `ensemble.cu` file and found that the `fixed_group` variable only affects the velocity advancement:
https://github.com/brucefan1983/GPUMD/blob/103b79ea98ec9d011fc555d1fcb5ea5f23eff1e6/src/integrate/ensemble.cu#L62-L70
As a result, the fixed groups should not affect the `operator_A`, which only does the position advancements. But as you can see, in the BAOAB Langevin thermostat, we do not measure the temperature immediately after the thermostat step (the `O` operator) but wait until the other two operators (one `operator_A` and one `operator_B`) are applied. This surely affects the temperature control but also improves the sampling accuracy in the conformational space. So, I would only advise those who are interested in conformational sampling to use this integrator.

```
[1] Leimkuhler, Benedict, and Charles Matthews. Applied Mathematics Research eXpress 2013.1 (2013): 34-56.
```

brucefan1983:
When a group is fixed, the positions for tha atoms in that group should not change, as in 
```
g_x[i] += vx * time_step;
      g_y[i] += vy * time_step;
      g_z[i] += vz * time_step;
```
where `vx, vy, yz` are set to zero. So you need to consider the fixed group in operator_A as well.

brucefan1983:
So "fixed group" means 
1) positions do not  change (frozen) and
2) velocities are zero (frozen)

The implementation in `ensemble.cu` looks like to considered velocities only, but it also considered position.

Lazemare:
Oh, I see, this is my fault! This must be a bug, I will fix this right now.

brucefan1983:
No worries, just want to know if the NEMD test can become better after this.

Lazemare:
Sure! After this bug is fixed, I will retest the NEMD runs and upload the results here.

Lazemare:
I have run the NEMD test again. But the results are identical. I think this is because the fixed group is not involved in the thermalized region, thus velocities of fixed atoms will not be modified by the `O` operator. And since the first operator applied to the phase vector is `B`, in which we already zeroed the velocities of these atoms, their velocities will keep at zero during the run.

brucefan1983:
This time, seems the temperature difference is correct, but the average temperature is shifted down.

Lazemare:
Yes, the average temperature is lower than the target temperature. I also observed similar phenomena when using the BAOAB Langevin integrator in OpenMM (provided by OpenMMTools).

brucefan1983:
So it is a common problem? Then should we disable this option for NEMD?

Lazemare:
I think this is appropriate. Because NEMD is about the sampling in the momentum space, there will be no good using the BAOAB splitting.

brucefan1983:
You can simply change the input part and removing the option of heat_bao. Other codes can be kept and we might want to test it in the future.

Lazemare:
Okay, I will undo my first commit in this PR.

Lazemare:
Now the `heat_bao` keyword has been disabled.

brucefan1983:
I merged the PR. Thanks for the great contribution!

brucefan1983:
I will update the manual (about `nvt_bao`) upon releasing GPUMD-v3.3.

Lazemare:
This is my pleasure! Thank you very much for your patient guidance!

PR #175, opened 2022-04-21
Make sure velocities in xyz.in are properly used for initialization

magnusrahm:
It seems like velocities as provided in the `xyz.in` file have no effect on an MD simulation. I think this is because velocities are read into `atom.cpu_velocity_per_atom` but never copied to the "GPU version" of the same array. This PR fixes this with one line, but I'm not sure it is the proper way to do it.

Here is a minimal working example with two atoms to test this:
[velocity-bug.tar.gz](https://github.com/brucefan1983/GPUMD/files/8532059/velocity-bug.tar.gz)

One of the two atoms is given a large velocity in the x direction. In the current master branch, it is quite clear that this does not affect the system. In this branch, that atom is properly shot onto the other.

### DIFF
--- a/src/model/read_xyz.cu
+++ b/src/model/read_xyz.cu
@@ -476,6 +476,7 @@ void allocate_memory_gpu(
   atom.position_per_atom.resize(N * 3);
   atom.position_per_atom.copy_from_host(atom.cpu_position_per_atom.data());
   atom.velocity_per_atom.resize(N * 3);
+  atom.velocity_per_atom.copy_from_host(atom.cpu_velocity_per_atom.data());
   atom.force_per_atom.resize(N * 3);
   atom.virial_per_atom.resize(N * 9);
   atom.potential_per_atom.resize(N);


brucefan1983:
You can call this a bug (based on your intuition), or not (based on the current document).

Here is the last paragraph in the document (https://gpumd.zheyongfan.org/index.php/The_velocity_keyword): 

If there are already velocity data in the xyz.in file, the initial temperature will not be used and the velocities will be initialized as those in the xyz.in file. In this case, you still need to write this keyword, otherwise the velocities will not be initialized at all (and will be undefined).

brucefan1983:
The velocity data are copied to GPU in line 265 of this file:

https://github.com/brucefan1983/GPUMD/blob/master/src/main_gpumd/velocity.cu

magnusrahm:
Oops, my bad, I missed that I had to use the `velocity` keyword (maybe a little counterintuitive?). Thanks for clarifying! I'll close this PR as this is not how it should be changed either way.

brucefan1983:
Let's keep it this way for the time being. The rules here are historical and kind of counter-intuitive. But perhaps some users have already gotten used to them.

PR #176, opened 2022-04-24
Implement NEP4, which combines NEP3 and message passing

brucefan1983:
Similar to #160 , but have definition for per-atom virial, which is crucial for MD applications. 

I made the following modifications:
  - Assume the `theta` matrix is an identity matrix (otherwise the code is unacceptably slow). With this simplification, NEP4 is only a couple of times slower than NEP3 with the same cutoff. I understand that the `theta` matrix is used to create new "features" from old ones, but I think that is the job of the hidden layer of the neural network applied later. I don't think it is really needed here. From my tests, the `theta` matrix is the source of the slowness. Removing it leads to speedup by a few times.
  - Assume the activation function `sigma` for message passing is an identity function (linear activation). I think `tanh` is not appropriate here after removing the `theta` matrix. I also normalized the descriptor after message passing, not the NEP3 descriptor.
  - Add atom type information into the messsage passing descriptor.
  
Based on the assumptions above, the site energy of atom `i` is simply
```
  U_i = NeuralNetworkFunction({ \sum_j d_ij fc(r_ij) * q^j_nu })
```
where `d_ij` is a scaling parameter depending on the atom types of `i` and `j`, `fc(r_ij)` is the cutoff function and `q^j_nu` is the nu-th component of the NEP3 descriptor for atom `j`. I believe this simple expression already contains the essential message passing information. The `d_ij` parameters are trainable.

I have created a test in the folder `nep4_test/` (this folder will be removed upon the completion of this PR). The training set consists of 100 azobenzene molecules (from revised MD17) and the testing set consists of 1000 molecules. I compare NEP4 agains NEP3 with the same cutoff. The `nep.in` files are:

```
type 3 C H N
version 3 # or version 4 if using NEP4
cutoff 3 3 # only using 3 Angstrom to test
n_max  8 8 # only the angular part is used in NEP4, but both are used in NEP3
l_max 4 # only using 3-body descriptors to test	
batch 100  # full batch, as there are only 100 structures in the training set
population 40 # enough for testing
generation 10000 # I restarted once, so I have 20000 steps in total
```

Here are the comparison between NEP4 with `d_ij=1` and NEP3:

image

We can see that using the same (small) cutoff, NEP4 can acheive higher training and testing accuracy. In this case, NEP4 is 2x slower than NEP3. 

Here are the comparison between NEP4 with trainable `d_ij` and NEP4 with fixed `d_ij=1`:
image

We can see that making `d_ij` trainable might be a good idea. This makes the atom types more distinguishable in the message-passing descriptor.

TODO: 
  - Might need also consider radial descriptor components with larger cutoff in NEP4, which is needed for describing inter-layer vdW interactions.
  - Implement the GPUMD part of NEP4 and validate force using finite difference.


src/main_nep/nep4.cu
```diff
@@ -174,22 +173,20 @@ NEP4::NEP4(char* input_dir, Parameters& para, int N, int N_times_max_NN_angular)
...
-  // ann
   ann.w0 = parameters;
   ann.b0 = ann.w0 + ann.num_neurons1 * ann.dim;
   ann.w1 = ann.b0 + ann.num_neurons1;
   ann.b1 = ann.w1 + ann.num_neurons1;
   ann.c = ann.b1 + 1;
-  // gnn
-  gnn.theta = parameters + ann.num_para;
+  ann.d =
+    ann.c + nep_para.num_types_sq * (nep_para.n_max_angular + 1) * (nep_para.basis_size_radial + 1);
```
elindgren, line 189: Have you re-added the radial descriptors, or is this only with angular components?

src/main_nep/nep4.cu
```diff
@@ -232,22 +229,11 @@ static void __global__ find_q_scaler(const int N, const float* g_q, float* g_q_s
...
-    }
-  }
-}
-
-// Apply the message passing aggregation between neighbors, A*q_theta
 static __global__ void apply_gnn_message_passing(
   const int N,
   const NEP4::Para para,
   const NEP4::ANN ann,
+  const int* __restrict__ g_type,
```
elindgren, line 236: I assume these are the atom types?

src/main_nep/nep4.cu
```diff
@@ -258,62 +244,45 @@ static __global__ void apply_gnn_message_passing(
...
+  const float* __restrict__ g_q_scaler,
+  float* g_pe,
+  float* g_dU_dq)
 {
   int n1 = threadIdx.x + blockIdx.x * blockDim.x;
   if (n1 < N) {
     float q[MAX_DIM] = {0.0f};
     for (int d = 0; d < ann.dim; ++d) {
-      q[d] = g_q[n1 + d * N];
+      q[d] = g_q[n1 + d * N] * g_q_scaler[d];
```
elindgren, line 279: Why did you re-enable the scaling of the descriptors? I thought it wasn't needed in NEP3, but maybe I misunderstood you.

src/main_nep/nep4.cu
```diff
@@ -258,62 +244,45 @@ static __global__ void apply_gnn_message_passing(
...
+      float q_out_nu = g_messages[n1 + nu * N];
       for (int j = 0; j < num_neighbors_of_n1; ++j) {
         int index = n1 + N * j;
         int n2 = g_NL[index];
+        int t2 = g_type[n2];
         float r12[3] = {g_x12[index], g_y12[index], g_z12[index]};
         float d12 = sqrt(r12[0] * r12[0] + r12[1] * r12[1] + r12[2] * r12[2]);
         float fc12;
         find_fc(para.rc_angular, para.rcinv_angular, d12, fc12);
+        fc12 *= ann.d[t1 * para.num_types + t2];
```
elindgren, line 259: and these are the parameters `dij` I suppose?

elindgren (commented):
Nicely done!

src/main_nep/nep4.cu
brucefan1983: Here the name "basis_size_radial" is historical and misleading. It is actually the "radial function basis size for angular descriptors". 

There are still angular descriptor only in NEP4. I am considering adding the radial part.

src/main_nep/nep4.cu
brucefan1983: Yes, the atom types (Chemical symbols) are mapped to some integers from 0 to `num_elements` - 1. In this example:
```
C = 0
H = 1
N = 2
```
according to this line in `nep.in`:
```
type 3 C H N
```

src/main_nep/nep4.cu
brucefan1983: Because I started this branch from master, which still has the scaling code. I suggested you remove it because I think it would be hard to figure out how to deal with the scaling for the derivatives. 

Now for my implementation, the scaling is for the message-passing descriptor, not the NEP3 descriptor and the scaling of the direvatives is as simple as before. So I simply kept it.

src/main_nep/nep4.cu
brucefan1983: Yes, these are the `num_types * num_types` `d_ij` parameters. I do not require the symmetry `d_ij = d_ji`.

src/main_nep/nep4.cu
elindgren: Alright makes sense :+1:

src/main_nep/nep4.cu
elindgren: Why do you not have the symmetry? `fc12` is symmetric, and `d_ij` should also be symmetric right? Wouldn't the potential violate permutational invariance if not `d_ij =  d_ji`?

brucefan1983:
Permutation symmetry does not require `d_ij = d_ji`. You can work out the details by playing with a few atoms. You can also see Eq. (3) and the related disussion in the paper:
https://arxiv.org/pdf/2203.08458.pdf

brucefan1983:
This does not seems to be a promising approach. I give up.

PR #179, opened 2022-04-30
Add examples using the carbon NEP potential

brucefan1983:
These examples are for the NEP3 paper.


hityingph (approved):
I have reviewed these examples. I changed some discriptions on the NEP3 potentials to keep consistent with each other. In addition to this, everything is okay.

PR #180, opened 2022-04-30
Fixed a bug in calling PLUMED for multiple times.

Lazemare:
This patch should fix the bug caused by the PLUMED interface when performing multiple runs in one `run.in` file.


brucefan1983:
Good. I did not used descructtors much in designing GPUMD. The changes look reasonable.

PR #181, opened 2022-05-01
Fixed the error on some units

hityingph:
I have fixed some errors on the units of temperature and VDOS in the NEP3 example of carbon.


brucefan1983:
perhaps you also need to update the Python scripts?

PR #182, opened 2022-05-01
remove two large files in an vasp2nep example

brucefan1983:
There are two files in this example that are too large (about 10 MB). Here I remove them to reduce the GPUMD package size.


Yanzhou-Wang (commented):
It is ok.

Yanzhou-Wang (approved):

PR #184, opened 2022-05-01
Add select_train for processing train.in or test.in

AmbroseWong:
Hi! I'm Yong Wang. I pushed my scripts to process and select structures from an initial train.in or test.in. This allows users to construct their dataset in more detail.


brucefan1983:
In the readme file
```
Usage:  
1. Select i-th to j-th strutures of the existing train.in file:  
python select_train.py existing-train.in i:j output-train.in
```
Perhaps it is better to make it clear that
1) the indexing 0-based or 1-based.
2) the j-th structure is included or not.

PR #185, opened 2022-05-01
Add select_train for processing train.in or test.in

AmbroseWong:
Add select_train scripts in tools/nep_related for processing and selecting structures from train.in or test.in. This allows users to construct datasets in more detail.


brucefan1983:
Thanks for the nice tools.

PR #187, opened 2022-05-04
Fix nep cpu bugs

brucefan1983:


brucefan1983:
The bugs occur in some special cases:
1) when descriptor is calculated before force calculation
2) when the second structure has smaller number of atoms the the previous one

elindgren (approved):
:+1: Looks good to me :)

brucefan1983:
thanks. hope you have not suffered from these bug yet.

elindgren:
> thanks. hope you have not suffered from these bugs yet.

I don't think I've encountered them :sunglasses:

PR #188, opened 2022-05-04
output latent space

brucefan1983:


brucefan1983:
@elindgren This so-called "latent space" is similar to the descriptor, which might be even better than the descripor for analyzing training data.  It reprents the state of the hidden layer (the last one, bu there is only one hidden layer in NEP) weighted by the weights.

elindgren (approved):
Nice! It would probably be very instructive to be able to study these as well. :+1:

PR #190, opened 2022-05-15
apply PBC before neighbor list update, not after

brucefan1983:
This is a bug in GPUMD that can potentially affect a simulation with the following conditions:
- The neighbor list is requested to be updated
- The box length in some direction is slightly smaller than the product of the cutoff distance (force cutoff  + skin) and an integer. For example, the cutoff is 9 Angstrom and the box length in some direction is 80.9 Angstrom, which is `delta` = 0.1 A smaller than 81 Angstrom. In principle, if `delta` < half of the skin distance (usually 1 A), it is potential to cause problems.
- The box is large enough such that GPUMD chooses to use the O(N) algorithm instead of the O(N^2) algorithm.

When this happens, atoms around x = 0 (if this is the problematic direction) will have chance to get fewer number of neighbors than the true value. 

To all users: if you think some of your simulations meet the conditions above, we suggest you check the results using the code after this PR.

### DIFF
--- a/src/model/neighbor.cu
+++ b/src/model/neighbor.cu
@@ -242,7 +242,10 @@ void Neighbor::find_neighbor(
   double* z = position_per_atom.data() + N * 2;
 
   if (is_first) {
+    // apply the PBC before constructing the neighbor list (this has been a bug)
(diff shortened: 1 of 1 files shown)


hityingph (approved):
Okay, it can be merged into the master verison.

PR #192, opened 2022-05-16
use mt19937 rng for BDP thermostat

brucefan1983:
The `ran1()` function in Bussi's original code seems to be too simple, leading to correlated results from differnet runs.

### DIFF
--- a/src/integrate/svr_utilities.cuh
+++ b/src/integrate/svr_utilities.cuh
@@ -16,61 +16,24 @@
 /*----------------------------------------------------------------------------80
 // The following functions are from Bussi's website
 // https://sites.google.com/site/giovannibussi/Research/algorithms
-// I have only added "static" in front of the functions,
-// without any other changes
+// I have only added "static" in front of the functions, and changed ran1()
+// to C++ calls
 // Reference:
 [1] G. Bussi et al. J. Chem. Phys. 126, 014101 (2007).
 ------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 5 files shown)


hityingph (approved):

brucefan1983:
In the figure below, old and new means the code before and after this PR. 1 and 2 refer to two independent runs (with different initial velocities). It can be seen that the two simulations are correlated using the old code.


image

src/integrate/svr_utilities.cuh
```diff
@@ -83,8 +46,10 @@ static double gasdev()
   }
 }
 
-static double gamdev(const int ia)
+static double gamdev(const int ia, std::mt19937& rng)
 {
+  std::uniform_real_distribution<double> rand1(0, 1);
```
Lazemare, line 51: Would a `static` declaration improve the performance?

Lazemare:
I think the code is good enough, expect the only one question I commented on above.

src/integrate/svr_utilities.cuh
brucefan1983: using static variables might make the function a little bit faster, but any CPU part is not a bottleneck in GPUMD. Using static variables are not good for possible future multi-threaded extension. So I do not use static variables at all.

Lazemare: ok, that makes sense :+1:

Lazemare (approved):

Issue #203, opened 2022-06-14
Use atom symbols for all potential models, not only for NEP models. 


brucefan1983:
This will make the code and interfaces cleaner.

brucefan1983:
This is not needed any more as the problem was related to ASE reading, and it has been solved in this PR: #204

PR #204, opened 2022-06-14
change species to numbers for non-NEP potentials

brucefan1983:


brucefan1983:
One of the motivations of  #203 
is to make the `dump.xyz` file readable by ASE and this PR should have solved it.

hityingph (approved):

PR #214, opened 2022-07-13
multiple GPUs for NEP

brucefan1983:
* Aim: add multi-GPU support for the NEP part, aiming for the following performance with eight 80-GB A100 within one node:
  - size: about 2e8 atoms 
  - speed: about 1e8 atom-step/second

* Related issue: #5 
  - I will not consider `MPI`, and I think it is better to restrict to a single node.


brucefan1983:
This try is not successful. Will do it again in another PR.

PR #223, opened 2022-07-25
Multi_GPU NEP_train

AmbroseWong:


brucefan1983:
In nep3.cu, Perhaps there is no need to use CUDA stream here. You can try to remove it and test if it is ok.

brucefan1983:
image

So here you want to consider the test set for calculating q_scaler?

AmbroseWong:
> image
> 
> So here you want to consider the test set for calculating q_scaler?

AmbroseWong:
> image
> 
> So here you want to consider the test set for calculating q_scaler?

I've removed the q_scaler to the Para class. Then I defined an array of  q_scaler to distribute it to each device and to make it dataset independent.

AmbroseWong:
> In nep3.cu, Perhaps there is no need to use CUDA stream here. You can try to remove it and test if it is ok.

I've tested a script and found that it's ok if don't use a stream per device. But this would cause the unavailability of the cudaMemcpyAsync when we want to copy data during a stream iteration. I think it's a good habit to create stream(s) when we try to use multi-GPUs.

PR #228, opened 2022-08-14
remove the triclinic keyword in model.xyz

brucefan1983:
* Let the code to check the box shape automatically at any time.
* Not finished yet. Many files need to be updated as the code needs to check if the box is triclinic before using it many in functions.
* Also need to modify the data structure, as currently box.cpu_h[] is used for both trilinic and orthogonal box. Need to use two arrays if we want to check the box shape automatically at any time.
* This is an update after PR #225, where I used a `triclinic` keyword in the second line of the extended XYZ file `model.xyz`. This seems to be not very conventional and I think it is better to remove it.

### DIFF
--- a/src/model/box.cu
+++ b/src/model/box.cu
@@ -123,3 +123,14 @@ bool Box::get_num_bins(const double rc, int num_bins[])
   }
   return use_ON2;
 }
+
+void Box::update_triclinic()
+{
+  if (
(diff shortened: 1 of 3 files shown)


brucefan1983:
After playing with the code for a while I realized this is a nontrivial task, as it involved many things, such as the various conditions for the NPT ensemble. I give up for now.

Issue #240, opened 2022-08-27
units for velocity in model.xyz


brucefan1983:
Need to check what are the units for velocity as produced by e.g., ASE.

Do this before releasing GPUMD-v3.4.

PR #243, opened 2022-08-27
Velocity units

brucefan1983:
* This PR can resolve this issue:  #240 
* Unify the units for velocity in all the input and output files, using `Angstrom/fs`.
* Warning messages are used to remind the reader about this change.
* The following manual pages are updated/created:
  * https://gpumd.zheyongfan.org/index.php/The_restart.out_output_file
  * https://gpumd.zheyongfan.org/index.php/The_model.xyz_input_file
  * https://gpumd.zheyongfan.org/index.php/The_velocity.out_output_file
  * https://gpumd.zheyongfan.org/index.php/The_restart.xyz_output_file
* I have not removed the `dump_velocity` command because it still contains an option that is not available in `dump_exyz`, i.e., choosing a group of atoms to dump. Perhaps we need to consider this option for `dump_exyz` in the future.
* All the regression tests passed except for the one with `dump_velocity`, which is expected. The reference data for that regression test are thus updated.


hityingph (approved):

erhart1:
This looks good.
I have only one question.
Didn't you plan to remove the `dump_velocity` keyword/functionality as well?

brucefan1983:
> This looks good. I have only one question. Didn't you plan to remove the `dump_velocity` keyword/functionality as well?

I plan to do it be not this time, as there is still something that can be done by `dump_velocity` but not `dump_exyz`, namely, the `group` option in `dump_velocity`.

elindgren (approved):

PR #247, opened 2022-08-30
hnemd for diffusive

liangzhixin-202169:
'ec' means Evans and  Cummings algorithm. There are 2 driving forces, one is heat force in PhysRevA.45.2370, which computes onsager coefficient Lqq and L1q,  the other is diffusive force in PhysRevA.35.5156, which computes onsager coefficient L1q and L11

### DIFF
--- a/src/force/force.cu
+++ b/src/force/force.cu
@@ -27,6 +27,7 @@ The driver class calculating force and related quantities.
 #include "tersoff1988.cuh"
 #include "tersoff1989.cuh"
 #include "tersoff_mini.cuh"
+#include "utilities/common.cuh"
 #include "utilities/error.cuh"
 #include "utilities/read_file.cuh"
 #include <vector>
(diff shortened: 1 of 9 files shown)


brucefan1983:
This is a great contribution. Here are some general comments:
* Remove any code that is commented out, as they are distracting and confusing
* Could I understand that HENMDEC only applies to two-element systems? I am actually not clear if the theory really cares about "elements" or just "molecules". For example, will the code treat H2O the same way as Ar-Kr fluid? In my opinion, water is one-component fluid, while Ar-Kr is two-component fluid.

liangzhixin-202169:
Yes, the algorithm just applies to two-element systems that two elements diffuse independently. I use 200GPa and 3000K water to test the code because it's superionic state, which means H can diffuse freely but O stay at its lattice, so I think H2O can be treat as two-component fluid.

brucefan1983:
Thanks, I will merge, but we will only document this part after we have more testing results and a publication.

Issue #237, opened 2022-08-26
nep_cpu is broken in master


brucefan1983:
Need to fix it before releasing GPUMD-v3.4.

brucefan1983:
Fixed in PR #253

PR #253, opened 2022-08-31
fix nep_cpu

brucefan1983:
* This PR fixes the broken `nep_cpu` as mentioned in this issue #237 
* Now it can accept both the old and the new `nep.txt` files.


brucefan1983:
Here are the testing results:

image

elindgren (approved):

Issue #254, opened 2022-09-01
use atom symbol for all potentials


brucefan1983:
This would be good and it just requires to modify the emperical potential files.

brucefan1983:
Fixed in PR #255

PR #255, opened 2022-09-01
Use atom symbols for all the potentials

brucefan1983:
* To resolve issue #254 
* Regression tests passed.
* Examples updated.
* Manual updated.


Yanzhou-Wang:
ok

erhart1:
Very good change 👍

PR #271, opened 2022-09-09
Updata run.cu

Jonsnow-willow:
Fix the bug that once the time step is reduced, it cannot be restored to the original data


brucefan1983:
Nice fix. Thanks!

PR #278, opened 2022-09-22
Parse stresses instead of virials in train/test.xyz

### LINKED ISSUE #275: Support for virials tag in extended xyz input
As of version 3.4 the input structures for the `nep` executable can be provided in extended xyz format, which is very useful.
In the currently supported format the virials can be provided via the `virial` tag in the second line. Unfortunately `ase` does not write the virials but the stresses.
One can of course write a custom extended xyz writer but it could be even easier to add support for the stresses as well.
The conversion is trivial as the stress tensor in the units written by `ase` only has to be multiplied by minus times the cell volume (in Å$^3$).
Hence, would it be possible to add support for parsing the stress tensor as well?
Afaict it would only require adding a block near line 198 in `main_nep/structure.cu`, no?

elindgren:
This PR adds functionality for reading stresses instead of virials in the tŕain/test.xyz. 
If
- only virials are defined -> virials will be used
- only stress is defined -> stress will be converted to virials and used
- both stress and virials are defined -> virials and stress will be checked for consistency.
    - If not consistent, an error will be thrown
    - If consistent, virials will be used
- If neither are defined, neither will be used.

ping @erhart1 

Fixes #275


brucefan1983:
Very nice! I will go through it. Have you tested the using your inputs?

brucefan1983:
Does ASE really have that minus sign? That is strange.

brucefan1983 (approved):

elindgren:
> Very nice! I will go through it. Have you tested the code using your inputs?

Yes I've tested it both manually and with input structures that we write through `calorine`. I had to increase the tolerance in stress/virials to around 1e-3 for it to work though. 

> Does ASE really have that minus sign? That is strange.

I believe it does, we have implemented it that way in our calculators at least. I'll double check before we merge this.

elindgren:
> Does ASE really have that minus sign? That is strange.

You were right, there should not be a minus sign (since both are stresses). I was thinking of the pressure.

brucefan1983:
perhaps you cannot see my comment on the code. Here is my comment:

You wrote:

both stress and virials are defined -> virials and stress will be checked for consistency.
If not consistent, an error will be thrown
If consistent, virials will be used

but the code  assinged stress to virial when both are present.

elindgren:
> perhaps you cannot see my comment on the code. Here is my comment:

I didn't see the review comment, but your completely right. I've fixed it in the latest commit,

> Does ASE really have that minus sign? That is strange.

I've triple-checked the minus-sign, and it should indeed be there. The virials as they are defined in `GPUMD` are proportional to $- r_{ij}f_{ij}$, as the per atom virials are defined as $v_i = \sum_{j > i} ) (-1) r_{ij} f_{ij}$. For example, in `nep3.cu`:

```c++
s_virial_xx -= r12[0] * f12[0];
```
The virial stress $\sigma$ on the other hand is proportional to $r_{ij}f_{ij}$. At $T=0$, we get $\sigma = \frac{1}{V} r_{ij} f_{ij}$. Thus there should be a minus sign when converting from stress to virials.

brucefan1983:
> > perhaps you cannot see my comment on the code. Here is my comment:
> 
> I didn't see the review comment, but your completely right. I've fixed it in the latest commit,
> 
> > Does ASE really have that minus sign? That is strange.
> 
> I've triple-checked the minus-sign, and it should indeed be there. The virials as they are defined in `GPUMD` are proportional to −rijfij, as the per atom virials are defined as vi=−∑j>irijfij. For example, in `nep3.cu`:
> 
> ```c++
> s_virial_xx -= r12[0] * f12[0];
> ```
> 
> The virial stress σ on the other hand is proportional to rijfij, see [here](https://en.wikipedia.org/wiki/Virial_stress). At T=0, we get σ=1V∑irijfij. Thus there should be a minus sign when converting from stress to virials.

ok if you get expected training results with the new stress item in `train.xyz`, then everygthing should be fine.

brucefan1983:
I don't have more comments and you can merge when you feel it's ready.

PR #287, opened 2022-10-21
Added first part of documentation in sphinx

erhart1:
This is the first part of the new sphinx based documentation.


brucefan1983:
Could you see my comments on the files? If not, I will just write here.

erhart1:
Where should I be able to see the comments?

brucefan1983:
for example, I have tried to commented like this:

image

brucefan1983:
Perhaps it's because you are trying to merging from outside. Similar situations have been encountered by others. 

Then we can only communicate by the comments here.

erhart1:
Sorry. The changes in the makefile were local and not meant to be included.
I reverted them in f1b8aec6.

erhart1:
I am actually not merging from outside but created a new branch directly in this repo and then created a PR from that one.

brucefan1983:
ok, then all the files are just new ones. you can tell me when you are done. if you think it is safe to merge I will do it. we can see what happens later.

erhart1:
This should be safe to merge.

erhart1:
It includes all changes we discussed today plus the output files and some smaller adjustments.
I could not get the direct push from GL to work so I manually compiled the changes in a new branch of your original repo.

brucefan1983:
so you mena it's ready to merge?

erhart1:
Yes

PR #291, opened 2022-10-22
Removed version selection from documentation

erhart1:
I removed the possibility to select different versions of the documentation for now since it does not work as long as there has not been a release that includes the documentation.
At the moment this breaks the CI.
Once the latter has happened we can reintroduce this possibility.


brucefan1983:
I don't think we ever need differnt versions. I only look forward, not backworkd :-)

erhart1:
That's the spirit. Sorry for the multiple PRs. Fixing the CI between two platforms is not entirely trivial.

PR #301, opened 2022-11-08
allow for longer systems for NEP with small box

brucefan1983:
Previously, it was not allowed to have a thickness < 2.5 radial cutoffs in a periodic direction and simultaneously a thickness > 5 radial cutoffs in another direction. The purpose of this is to prevent the user from using a non-optimal box (because in this case the "small-box algorithm" will be used and there are still many atoms, which can make the code slow), but Erik hoped to have more freedom for the box dimensions, for the active-learning purpose. Now I relax it a lot.

### DIFF
--- a/src/force/nep3.cu
+++ b/src/force/nep3.cu
@@ -986,11 +986,11 @@ static bool get_expanded_box(const double rc, const Box& box, NEP3::ExpandedBox&
   }
 
   if (is_small_box) {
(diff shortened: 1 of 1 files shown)


erhart1:
This looks good to me.
@elindgren do you want to test this before merging?

brucefan1983:
it's so simple fix. I think no need to test. I will merge soon, to let Erik to use asap.

PR #312, opened 2022-11-15
Updated makefile to avoid unnecessary warnings and errors

erhart1:
This PR introduces four small changes to the makefile that prevent several warnings and errors during the compilation.
* The minimal compute capability is changed from `sm_35` (Kepler) to `sm_50` (Maxwell). This prevents a deprecation warning. The [Kepler architecture](https://en.wikipedia.org/wiki/Kepler_(microarchitecture)) is from 2012 and hence very outdated. Even the [Maxwell architecture (2014)](https://en.wikipedia.org/wiki/Maxwell_(microarchitecture)) is rather old and likely to become deprecated relatively soon. See [here](https://arnon.dk/matching-sm-architectures-arch-and-gencode-for-various-nvidia-cards/) for a translation table between `sm_XX` and architecture.
* The C++ standard was switched from C++11 to C++14 (`-std=c++14`) since the former also causes a deprecation warning.
* A stray tab was removed, which caused my emacs to act up.
* The `rm` command used by `make clean` now uses the `-f` flag such that it does not throw an error when `make clean` is run without `gpumd` and `nep` present. This otherwise triggers an error preventing the rather common `make clean && make` sequence to run (i.e., `make` is run on the condition that `make clean` passes).

### DIFF
--- a/src/makefile
+++ b/src/makefile
@@ -1,6 +1,6 @@
 ###########################################################
 # Note: 
-# 1) You can modify -arch=sm_35 according to 
+# 1) You can modify -arch=sm_50 according to 
 #    your GPU architecture.
 # 2) For Windows systems, if you get errors like 
 #    c1xx : fatal error C1083: Cannot open source file: ...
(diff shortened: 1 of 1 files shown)


erhart1:
ping @brucefan1983

brucefan1983:
Great. I was also thinking to change `sm_35` to something new. Perhaps it is better to change to `sm_60` (Pascal), as Maxwell will also be deprecated too. Pascal has native double-precision atomic function, which is much faster than a hand-made one for pre-Pascal.

erhart1:
I updated to `sm_60`.

erhart1:
Please merge whenever you're happy.

PR #314, opened 2022-11-15
Pages job allowed to run independent of schedule

erhart1:
Removed limitation to run pages job only on schedule.


brucefan1983:
ready to merge?

Issue #295, opened 2022-10-27
Increse weight for shearing virial


brucefan1983:
Shearing (off-diagonal) virial components have relatively small values compared to diagonal ones (about 10x smaller) and with a uniform weight, one can get inaccurate predictions for shearing stress when it is very small. Sometimes even the sign is wrong.

Someone has tested that this can make C44 (and similar elastic constants) much better. A PR is needed to make this available to all users.

brucefan1983:
solved in PR #320

PR #320, opened 2022-11-18
allow for choosing shear virial weight

brucefan1983:
In most cases, the shear virials/pressures in the training data are much smaller than the diagonal ones and it is might be beneficial to increase the weight for the shear part. Now one can use
```
lambda_shear <value>
```
in `nep.in` to set such a weight. Someone tested that a value of 5 can make the prediction of `C_44` much more accurate in some material.

To solve #295


erhart1:
Looks good to me.
I would probably rather include more specifically deformed configurations but this seems works for some.
Do you still want to add a page to the documentation?

brucefan1983:
Yes, I will update doc.

brucefan1983:
> Looks good to me. I would probably rather include more specifically deformed configurations but this seems works for some. Do you still want to add a page to the documentation?

In some cases it is even insufficient to include specifically deformed configurations according to tests by Zhiqiang (I think he has contact you for entory maximization). The problem is the shear virials/moduli are much smaller than diagonal ones in some materials (one order of magnitude smaller).

brucefan1983:
I messed up by doing something accidently in master...
perhaps need to start this PR over

brucefan1983:
I fixed the conflict and it is ready to merge to master. 

documentation has been directly committed to master.

PR #322, opened 2022-11-20
Fixed syntax error in the documentation of lambda_shear

erhart1:
This fixes a syntax error in the documentation of the `lambda_shear` keyword that caused the compilation of the user guide to fail.
It also includes the addition of `gpumd` and `nep` to `.gitignore` to avoid these executable files being shown under "untracked files".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ src/nep*
 local
 public*
 \#*
+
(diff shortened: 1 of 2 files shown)


brucefan1983:
Thanks for fixing these.

Issue #242, opened 2022-08-27
compute_msd


brucefan1983:
Although MSD is mathematically equivalent to VAC, adding a `compute_msd` command will be very helpful, as it does not require a small time interval.

brucefan1983:
solved in PR #324

PR #324, opened 2022-11-22
implement MSD calculations

brucefan1983:
To solve issue #242 

* Added the following keyword for `run.in`:
```
compute_msd <sample_interval> <num_correlation_steps>
```

* An example in `examples/diffusion` is added. The `run.in` reads:
```
potential   ../../potentials/nep/Si_2022_NEP2.txt
velocity    2500

time_step   2

ensemble    npt_scr 2500 2500 100 0 50 1000
dump_exyz   10000 0 0
dump_thermo 100
run         10000

ensemble    npt_scr 1800 1800 100 0 50 1000
dump_exyz   10000 0 0
dump_thermo 100
run         10000

ensemble    nve
dump_thermo 100
dump_exyz   100000 0 0
compute_sdc 1 2000
compute_msd 1 2000
run         20000
```

* I tested silicon at 1800 K using the NEP model in `potentials/nep/`. The figure below shows that the SDC calculated from MSD is equivalent to that from VAC (they are rigorously equivalent theoretically but it needs a small sample interval to achieve good agreement numerically)

image

@Kick-H You can comment on the PR as you have worked on this before.


ZKC19940412:
Hi, I have also tested both ways of computing diffusion coefficients and I have some questions if I may. First, diffusion coefficients computed with MSD are meaningful if the MSD is computed from an unwrapped trajectories and I would assume this is the case here right? Second, diffusion coefficients computed with MSD is typically done by fitting the slope of the linear region of the MSD, which means the diffusion coefficients should be a scalar. I wonder why we see convergence behavior for SDC with MSD in figure c. Maybe I am misunderstanding something here? 

Thank you so much for asking my questions.

Best, 
Zekun

brucefan1983:
> rst, diffusion coefficients computed with MSD are meaningful if the MSD is computed from an unwrapped trajectories and I would assume this is the case here right? Second, diffusion coefficients computed with MSD is typically done by fitting the slope of the linear region of the MSD, which means the diffusion coefficients should be a scalar. I wonder why we see convergence behavior for SDC with MSD in figure c. Maybe I am misunderstanding

1) Yes, I used unwrapped trajectory to calculate MSD.

2) It is a misconception in the community that one must use a linear fit to MSD somewhere to get a diffusion coefficient. Mathematically, we have the following identity:

$$ 
\frac{1}{2} \frac{d}{dt} \Delta x(t) = \int_0^t C_{xx}(\tau) d\tau
$$

In figure c, the two curves are from the two sides of the equation above, which are totally equivalent.

For a proof of the equation above (the equivalence of Green-Kubo and Einstein), see, e.g., my notes (in Chinese): https://github.com/brucefan1983/Molecular-Dynamics-Simulation/blob/master/chapter-5-transport/GK_vs_Einstein.pdf

PR #326, opened 2022-11-24
Added MSD as abbreviation to glossary

erhart1:
This PR fixes a bug in the documentation.


brucefan1983:
Thanks. I am still not very familiar with the grammars for the doc.

erhart1:
If you want to make sure that the documentation compiles correctly you run the following command
```shell
sphinx-build -W doc public
```
in the root directory of the repository.

PR #334, opened 2022-12-01
allow for up to 103 elements for nep

brucefan1983:
Needed for crazy guys to train a NEP for many elements.


elindgren (approved):
Looks good to me. :fire:

PR #337, opened 2022-12-04
Improve nep4

brucefan1983:
This is NEP5, a version that complements NEP4. NEP4 will not work when there are only pure subsystems (A, B, C, ...), but NEP5 can work even in this case. However, it seems there is no advantage of NEP5 over NEP4 when there are compound subsystems (such as AB, AC, BC, ABC, ...). Because in practice one should not only have pure subsystems, we should not need NEP5 at all. So I will probably abandom this PR, but I would like to wait for some converged training results...


brucefan1983:
Conclusion: this NEP5 has no advantages over NEP4. So I close this PR.

PR #339, opened 2022-12-05
fix a CUDA kernel configuration bug

brucefan1983:


brucefan1983:
Someone found that PLUMED cannot be run with more than 1024 atoms, and it turned out to be an error here.

PR #356, opened 2023-01-05
not require abs(c)>0.1

brucefan1983:
In my NEP2 paper, I reauired that abs(c)>0.1 for the descriptor parameters c. Now I think this restriction is not needed. Without this restriction I have observed L1 regularization can select features better and the regression accuracity is usually higher.


elindgren (approved):
:rocket:

Issue #357, opened 2023-01-07
A way to choose regularization parameters automatically?  


tamaswells:
I don't know the order of magnitude of RMSE before training. So I need to tune the regularization paramerters via trial-and-error  Is there a way o choose regularization parameters automatically?  Many thanks.

shdchen:
Perhaps, the regularization parameters could be determined automatically by the RMSE (weighted sum of the energy, force, and virial RMSEs) on-the-fly. It is under testing. Thanks.

brucefan1983:
done in #358

PR #358, opened 2023-01-07
improve polarizability model and use auto reg

brucefan1983:
* Improve the polarizability model by using different NN parameters for the scalar and the tensor parts. This is crucial for small molecules.
* Change the default values of `lambda_1` and `lambda_2` to -1, which means they are automatically calculated as the physical RMSE (weighted sum of the energy, force, and virial RMSEs) of the elite (the best solution) at every generation. Based on tests, this is quite opitimal. For example, when the physical RMSE is only 0.01 (in some units), if we still use the previous default `lambda_1=lambda_2=0.05`, the regularization will be too strong to cause underfitting. Below is a test showing that this automatic scheme works fine (the fixed-reg case used `lambda_1=lambda_2=0.01`, which was found by trial and error):

image

* Solve #357


doc/nep/input_parameters/lambda_2.rst
```diff
@@ -10,9 +10,6 @@ The syntax is::
...
 
-Here, :attr:`<weight>` represents :math:`\lambda_2`, which must satisfy :math:`\lambda_2 \geq 0` and defaults to :math:`\lambda_2 = 0.05`.
-It is often beneficial for model stability to use a strong regularization with :math:`\lambda_2 = 0.1`.
-
-In practice, one can first estimate the weighted sum of the loss terms for the target quantities upon convergence and then 
-set :math:`\lambda_2` to that value. For example, if one estimates the weighted sum of the loss terms for energy, force, 
-and virial to be :math:`0.01 + 0.1 + 0.01`, then it is reasonable to set :math:`\lambda_2 = 0.12`.
+Here, :attr:`<weight>` represents :math:`\lambda_2`, which can be set to any non-negative values. 
+However, the default value is :math:`\lambda_2 = -1`, which means that it will be automatically determined based on the actual loss terms for the physical quantities. 
+We have tested that the default is quite optimal, but the users can also test manually chosen values.
```
elindgren, line 15: In which situations does this default not work as optimally? When the RMSE are very different for energies/forces/virials? Or when the RMSE becomes too small?

If you have any such empiric understanding, it would perhaps be good to add that to the docs, in order to help users identify if they need to set it manually.

src/main_nep/nep3.cu
```diff
@@ -308,7 +308,25 @@ void NEP3::update_potential(float* parameters, ANN& ann)
...
+      ann.b0_pol[t] = pointer;
+      pointer += ann.num_neurons1;
+      ann.w1_pol[t] = pointer;
+      pointer += ann.num_neurons1;
+    }
+    ann.b1_pol = pointer;
+    pointer += 1;
+  }
+
+  ann.c = pointer;
```
elindgren, line 329: Is there a written description of the polarizability model? Is it the same neural network architecture as in regular NEP, or is it modified in order to be equivariant? I'm curious on how it's implemented :)

src/main_nep/snes.cu
```diff
@@ -169,15 +169,34 @@ void SNES::create_population(Parameters& para)
 
 void SNES::regularize(Parameters& para)
 {
+  float lambda_1 = para.lambda_1;
+  float lambda_2 = para.lambda_2;
+  if (para.lambda_1 < 0.0f || para.lambda_2 < 0.0f) {
+    float auto_reg = 1.0e30f;
+    for (int p = 0; p < population_size; ++p) {
+      float temp = fitness[p + 3 * population_size] + fitness[p + 4 * population_size] +
+                   fitness[p + 5 * population_size];
```
elindgren, line 178: Does this mean that energies, forces and virials are weighted equally?

doc/nep/input_parameters/lambda_2.rst
brucefan1983: Yes,  you know that the NN parameters usually have values of the order of 1 or smaller, like 0.5. Then the L1 loss term will have a value of about `0.5*lambda_1`. Then if the weigthed sum of the energy, force, and virial RMSEs is only 0.001, the default value of `\lambda_1=\lambda_2=0.05` will make the regularization too strong, leading to underfitting. On the other hand, if the weigthed sum of the energy, force, and virial RMSEs is 1, the default value of `\lambda_1=\lambda_2=0.05` will make the regularization too weak, leading to underfitting.

I will add back the comments in the doc.

src/main_nep/nep3.cu
brucefan1983: This is still kind of under development (a paper is being prepared) and will be documented in GPUMD-3.7 I think. 

Basically, it only involves very minor modifications of the NEP model of potential energy, to get a model for polarizability as well as dipole.

Let's discuss this via email for more technical details.

src/main_nep/snes.cu
brucefan1983: the weights have already been included in this `fitness` array. You can have a look of the function `compute` in `fitness.cu`.

PR #359, opened 2023-01-08
Fixed a small bug in reading VASP OUTCAR

shdchen:
Fixed a bug that did not read the structure with lattice vector in a format like the snapshot below.

latt


Yanzhou-Wang:
Simple and good modification.

Yanzhou-Wang (approved):

brucefan1983:
nice fix, thanks

brucefan1983 (approved):

PR #360, opened 2023-01-09
Docs multiple nep

elindgren:
@brucefan1983 I've added docs for `dump_observer`, I think. How do I compile the docs locally to check that it's done correctly?


brucefan1983:
> @brucefan1983 I've added docs for `dump_observer`, I think. How do I compile the docs locally to check that it's done correctly?

Paul is the expert on this :-)

brucefan1983:
You can merge when you feel it is ready!

PR #361, opened 2023-01-09
Multiple loss

brucefan1983:
This PR makes a big improvement for NEP training with many elements, where the training data set contains systems with different compositions, such as a training data set with 20 elements and with 20 elementary subsets, 20*19/2 binary subsets, etc. In this case, it is benificial to define different loss terms for different subsets of parameters. This way, each subset of parameters will have a clear goal to evolve!

* Usage is not affected. The user will only experience faster convergence. The converged accuracy might be not affected. What improved is the convergence rate.
* A test for potential training in a 5-element system containing pure elements only (very significant improvement for NEP4 and also some improvement for NEP2 and NEP3):
image
* A test for polarizability training in a small HCNOSCl dataset containing compounds only (less significant improvement, which is expected):
image

### DIFF
--- a/src/main_nep/dataset.cu
+++ b/src/main_nep/dataset.cu
@@ -620,10 +620,9 @@ std::vector<float> Dataset::get_rmse_virial(Parameters& para, const bool use_wei
     }
   }
 
-  int num_components = (para.train_mode == 1) ? 3 : 6;
   for (int t = 0; t <= para.num_types; ++t) {
     if (count_array[t] > 0) {
-      rmse_array[t] = sqrt(rmse_array[t] / (count_array[t] * num_components));
+      rmse_array[t] = sqrt(rmse_array[t] / count_array[t]);
     }
   }
   return rmse_array;
--- a/src/main_nep/dataset.cu
+++ b/src/main_nep/dataset.cu
@@ -570,7 +570,8 @@ std::vector<float> Dataset::get_rmse_virial(Parameters& para, const bool use_wei
   int mem = sizeof(float) * Nc;
   const int block_size = 256;
 
-  float shear_weight = (para.train_mode != 1) ? para.lambda_shear * para.lambda_shear : 0.0f;
+  float shear_weight =
+    (para.train_mode != 1) ? (use_weight ? para.lambda_shear * para.lambda_shear : 1.0f) : 0.0f;
   gpu_sum_virial_error<<<Nc, block_size, sizeof(float) * block_size * 6>>>(
     N, shear_weight, Na.data(), Na_sum.data(), virial.data(), virial_ref_gpu.data(),
     error_gpu.data());
--- a/src/main_nep/dataset.cu
+++ b/src/main_nep/dataset.cu
@@ -63,6 +63,17 @@ void Dataset::copy_structures(std::vector<Structure>& structures_input, int n1, 
   }
 }
 
+void Dataset::find_has_type(Parameters& para)
+{
+  has_type.resize((para.num_types + 1) * Nc, false);
+  for (int n = 0; n < Nc; ++n) {
+    has_type[para.num_types * Nc + n] = true;
+    for (int na = 0; na < structures[n].num_atom; ++na) {
+      has_type[structures[n].type[na] * Nc + n] = true;
+    }
+  }
+}
+
 void Dataset::find_Na(Parameters& para)
 {
   Na_cpu.resize(Nc);
@@ -275,6 +286,7 @@ void Dataset::construct(
 {
   CHECK(cudaSetDevice(device_id));
   copy_structures(structures_input, n1, n2);
+  find_has_type(para);
   error_cpu.resize(Nc);
   error_gpu.resize(Nc);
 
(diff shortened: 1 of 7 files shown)


src/main_nep/dataset.cu
```diff
@@ -354,15 +366,25 @@ float Dataset::get_rmse_force(Parameters& para, const bool use_weight, int devic
...
+        rmse_array[t] += rmse_temp;
+        count_array[t] += Na_cpu[n];
+      }
     }
   }
-  return sqrt(error_sum / (N * 3));
+
+  for (int t = 0; t <= para.num_types; ++t) {
+    if (count_array[t] > 0) {
+      rmse_array[t] = sqrt(rmse_array[t] / (count_array[t] * 3));
```
elindgren, line 384: What does `count_array` contain, and why is it multiplied by a factor of 3? Three cartesian directions?

src/main_nep/dataset.cu
```diff
@@ -63,6 +63,17 @@ void Dataset::copy_structures(std::vector<Structure>& structures_input, int n1,
...
+{
+  has_type.resize((para.num_types + 1) * Nc, false);
+  for (int n = 0; n < Nc; ++n) {
+    has_type[para.num_types * Nc + n] = true;
+    for (int na = 0; na < structures[n].num_atom; ++na) {
+      has_type[structures[n].type[na] * Nc + n] = true;
+    }
+  }
+}
+
```
elindgren, line 76: Does this get the number of species? Are the losses then computed with one loss per species?

src/main_nep/dataset.cu
```diff
@@ -460,42 +486,48 @@ float Dataset::get_rmse_energy(
...
-      error_ave += total_weight * error_cpu[n];
+      float rmse_temp = use_weight ? weight_cpu[n] * weight_cpu[n] * error_cpu[n] : error_cpu[n];
+      for (int t = 0; t < para.num_types + 1; ++t) {
+        if (has_type[t * Nc + n]) {
+          rmse_array[t] += rmse_temp;
+          ++count_array[t];
+        }
+      }
     }
   }
```
elindgren, line 532: This block of code seems to be repeated almost exactly three times, with the difference being in gpu_sum_pe_error. Replace with a loop?

brucefan1983: good point, can use a loop or even combine the three kernels into one. I will try it out.

src/main_nep/dataset.cu
brucefan1983: `has_type` can be understood as a 2D array `has_type[num_types+1][num_structures]`. Suppose there are 3 structures, and the first only containes element `A` (type 0), the second only `B` (type 1), the third only `AB`, then the array will be calculated to be:

```
true  false true
false true true
true true true
```

The meaning of `has_type[t][n]` is "the structure n contains type t". The last column is always `true` because each structure contains "some type", which will contribute to the total loss.

src/main_nep/dataset.cu
brucefan1983: `count_array[t]` here is the total number atoms from the structures that contain type `t` atom and  yes, 3 refers to the three cartesian directions.

src/main_nep/dataset.cu
elindgren: Ah I see. 
Can you reiterate the part about  "some type"? Why is it necessary? Does it refer to a part of the loss that is independent of type?

brucefan1983: that is the original total loss, which is required for e.g. reporing the RMES in loss.out.

brucefan1983: that is, apart from the original total loss, now I calculate extra `num_types` losses, which are for A element, B element, C element, ...

brucefan1983: for the loss of A element, I consider system A (pure element), AB, AC, AD, ..., (binary alloys), ABC, ABC, ABD, ACD, ...

brucefan1983: To sum up, there are now `num_types + 1` loss function, one is the **global** loss that is used to optimize some global parameters and report the RMSE on screen and loos.out, and the other `num_types` are **local** ones that are used to optimize the different subsets of the parameters exclusively related to each element.

elindgren: Makes sense, thanks!

src/main_nep/dataset.cu
elindgren: Glad to hear that! :speedboat:

src/main_nep/snes.cu
```diff
@@ -174,12 +249,14 @@ void SNES::regularize(Parameters& para)
...
-      float temp = fitness[p + 3 * population_size] + fitness[p + 4 * population_size] +
-                   fitness[p + 5 * population_size];
+      float temp = fitness[p + (6 * para.num_types + 3) * population_size] +
+                   fitness[p + (6 * para.num_types + 4) * population_size] +
+                   fitness[p + (6 * para.num_types + 5) * population_size];
       if (auto_reg > temp) {
         auto_reg = temp;
       }
     }
+    auto_reg *= 0.4f;
```
shdchen, line 259: Not sure if 0.4 is good different systems. Let's do some more testings on this. (I'm running a couple of testings with different coefficients, which should be converged soon.)

brucefan1983: as we checked, 0.4 should be a good value.

PR #371, opened 2023-02-07
Dump only potential energies in dump_observer

elindgren:
For some applications writing the full `exyz` files is unnecessary. In those cases one would only want to dump e.g. the potential energy in a file similar to `thermo.out`. 

TODO
- [x] Recompute thermo for every potential
- [x] Dump only potential energy to new file
- [x] Add check to ensure all potentials have the same ordering of types
- [x] Add tests 
- [x] Update documentation


elindgren:
I also noticed an issue where the `thermo` properties where not recomputed for every potential after the energies, forces and virials had been computed. I've solved it like this: 

https://github.com/brucefan1983/GPUMD/blob/b43065bbf921c8fc591319040240a2bafafa09d2/src/measure/dump_observer.cu#L175-L177

@brucefan1983 Do you think this makes sense?

elindgren:
ping @erhart1

brucefan1983:
> I also noticed an issue where the `thermo` properties where not recomputed for every potential after the energies, forces and virials had been computed. I've solved it like this:
> 
> https://github.com/brucefan1983/GPUMD/blob/b43065bbf921c8fc591319040240a2bafafa09d2/src/measure/dump_observer.cu#L175-L177
> 
> @brucefan1983 Do you think this makes sense?

It looks good.

I have tried to make the following parts independent of each other:

* Force
* Integration
* Measure

But I think it would be too much work to ask you to strictly follow this convention. I can try to refactor the code in the future, perhaps by putting some CUDA kernels into a common module.

elindgren:
@brucefan1983 I have encountered a bug in `dump_observer`, which has to do with `atom.type`. If I supply two potentials with the same species, but the species in different order, then `atom.type` will only match one of the potentials. I see two options for fixing this bug:

1. Modify `nep3.cu` (or `potential.cu`) to include a GPU vector called `type`. This vector will be populated when each potential is read. Thus, one would not need to supply `atoms.type` to `compute`. 
2. Update `atoms.type` to suit the potential. This would potentially require rewriting the gpu vector `atoms.type` for every timestep and every potential (if dumping every timestep). 

I think that 1. is the more elegant choice, but I don't know if it has any implications elsewhere in the code. What do you think?

https://github.com/brucefan1983/GPUMD/blob/b43065bbf921c8fc591319040240a2bafafa09d2/src/measure/dump_observer.cu#L172-L173

brucefan1983:
Without a strong reason for not requiring this, I would prefer to keep the code unchanged, just make things simpler.

elindgren:
> Should we require that the different NEP models have the same order of atom types?

That is probably the simplest solution. The most natural would be for the potentials to keep track of the order themselves, but in lieu of that enforcing the same order makes more sense.

> Without a strong reason for not requiring this, I would prefer to keep the code unchanged, just make things simpler.

Understandable, I would not want to change every place where `compute` is called, sounds error prone :sweat_smile: 

@erhart1 @NicklasOsterbacka thoughts?

erhart1:
I can see the benefits of requiring models to have the same ordering.
This has the disadvantage of shifting the complication of reordering the parameters elsewhere, which I think we can handle though.

NicklasOsterbacka:
I agree with Paul. While it would be easier for our particular use case, I think it's easier to figure out how to reorder the NEP parameters than refactoring GPUMD here.

elindgren:
> Without a strong reason for not requiring this, I would prefer to keep the code unchanged, just make things simpler.

> I can see the benefits of requiring models to have the same ordering.

> I agree with Paul. While it would be easier for our particular use case, I think it's easier to figure out how to reorder the NEP parameters than refactoring GPUMD here.

Agreed, I'll implement a check to assert that all supplied potentials have the same ordering of types then :+1:

elindgren:
@brucefan1983 This one is ready for review :)

brucefan1983:
Thanks, I will check it during the weekend. There is another PR as you may
noticed, about more flexible ZBL. I hope there is no conflict.

Zheyong

brucefan1983:
The code looks great and I have only one comment: The examples in `tests\gpumd\dump_observer` take too much disk size (about 28 MB). I suggest use smaller problems to reduce the disk size (by one order of magnitude). You may not know how slow it could be to download the package in contries other than Sweden :-) We must keep the package size under control.

elindgren:
> The code looks great and I have only one comment: The examples in `tests\gpumd\dump_observer` take too much disk size (about 28 MB). I suggest use smaller problems to reduce the disk size (by one order of magnitude). You may not know how slow it could be to download the package in contries other than Sweden :-) We must keep the package size under control.

Done, I managed to reduce it to 816KB. Merging :)

PR #375, opened 2023-02-15
An update on zbl

Jonsnow-willow:
Added flexible ZBL to NEP


brucefan1983:
looks great. Thanks for this nice extention.

PR #386, opened 2023-02-28
a special batching method

brucefan1983:
For testing only.


brucefan1983:
This is not the correct way to go. So forget about it.

PR #388, opened 2023-03-03
speed up full-batch training by 15%

brucefan1983:
When using full-batch training, the neighbor list does not need to be repeately built, which can speed up training by at least 15%。


shdchen (approved):
Nice!

erhart1:
Sweet

Issue #79, opened 2021-01-14
Implement path-integral MD (PIMD) methods


brucefan1983:
Implement one or more PIMD methods, such as ring polymer MD. This should be done at some point.

brucefan1983:
The most useful paper I read on this topic: [Efficient stochastic thermostatting of path integral molecular dynamics](https://doi.org/10.1063/1.3489925)

brucefan1983:
I have started to implement PIMD/RPMD, but it seems to be a very demanding work. I don't want to have a big long-lived PR, so I will use a few small-scale PRs to achieve the goal.

PR #393, opened 2023-03-12
Implement PIMD and RPMD

brucefan1983:
# This PR aims to implement PIMD and RPMD

- To solve issue #79 
- Only consider NVT (PIMD) and NVE (RPMD) here, leaving NPT for future work.
- Usage: 
```
ensemble nvt_pimd <temperature> <tau> <step_within_pimd> <num_beads>
```
- `temperature` is the target temperature.
- `tau` is the coupling time (inverse of friction) in the local Langevin thermostat.
- `step_within_pimd` is the step (during this run) starting from which to switch off the thermostats for the centroid modes. For example, if this run has 20 000 steps, and `step_within_pimd` is 10 000, it means that the centroid modes will be thermostatted from step 0 to 9 999 (this is called PIMD) and not thermostatted from step 10 000 to 19 999 (this is called thermostatted ring-polymer MD, or TRPMD).
- `num_beads` is the number of beads, which must even numbers from 2 to 128.
- Full documentation will be left for another PR.

### DIFF
--- a/src/integrate/ensemble.cuh
+++ b/src/integrate/ensemble.cuh
@@ -14,6 +14,7 @@
 */
 
 #pragma once
+#include "model/atom.cuh"
 #include "model/box.cuh"
 #include "model/group.cuh"
 #include "utilities/gpu_vector.cuh"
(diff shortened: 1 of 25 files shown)


erhart1:
Do you want to add a documentation page already now or test the feature a bit more first?

brucefan1983:
> Do you want to add a documentation page already now or test the feature a bit more first?

Yes, I will add a documentation within the `ensemble` keyword, but this PR has not been fully tested. I am inviting people to test and also discuss the format of the keyword. I am still working on implementing some "estimators" such as those for kinetic energy and virial. For heat current, we even need to create some, as I have not seen a rigorously derived one.

brucefan1983:
We need to make a few tests to confirm the correctness of the code. I am considering the folloiwng:
1) RDF of O-H pairs in water, which is wrong even for an NEP model trained recently.
2) heat capacity of an arbitry matter. This should be the same as those calcualted by quantum correction based on DOS.

shdchen:
> 

To get the RDF correct, probably we would need the position information from all beads. Maybe we can consider adding a new command "dump_position_beads"?

brucefan1983:
> > 
> 
> To get the RDF correct, probably we would need the position information from all beads. Maybe we can consider adding a new command "dump_position_beads"?

might be!

brucefan1983:
I feel that I need merge this PR now and continue to do some improvements in the future. This PR does not affect existing features.

PR #395, opened 2023-03-21
Fixed a bug in the PLUMED interface.

initqp:
This fixes the bug caused by the different definitions of the virial in PLUMED and GPUMD. After this patch, the code could indeed pass the [virial contribution test](https://www.plumed.org/doc-v2.8/developer-doc/html/_how_to_plumed_your_m_d.html).


brucefan1983:
Good fix.

PR #396, opened 2023-03-21
Implement active learning

elindgren:
### What
Implement committee active learning in GPUMD.

**Requirements**
1. Supply a committee of `M` NEP potentials. The main potential used to run MD will be the first potential. Uncertainties will be computed over all `M` models.
2. At an interval `l` time steps, evaluate all potentials.
3. Compute the variance in predicted energy $e_i$ and forces $f_{ik}$ for atom $i$ and Cartesian direction $k$ over all models, $\sigma_{e,i}$ and $\sigma_{f,ik}$.
4. This iteration: Compute the max magnitude of **force** uncertainties, ignoring energy uncertainties atm since forces are more important for MD.
    - Compute $\sigma_f = max_i\left( \sqrt{\sum_k\sigma_{f,ik}^2} \right)$
5. If $\sigma_f > \delta$, where $\delta$ is the cutoff frequency, write the current structure to a file named `active.xyz`, in append mode.
6. The uncertainty should also be written in the exyz-format to `active.xyz`.
7. Every `l` timesteps, write $\sigma_f$ to a file named `active.out` regardless of if it exceeds $\delta$ or not.

### Uncertainty estimate

Let $f_{i,k}$ be the $k\in{x,y,z}$ Cartesian direction component of the force on atom $i$. Assume that $f_{i,k}$ is a normally distributed random variable, and that all force components are independently distributed, $f_{i,k} \sim \mathcal{N}\left( \mu_{i, k, true}, \sigma^2_{i,k, true} \right)$. Furthermore, let $[f_{i,k}^m]$ be the $M$ observations of $f_{i,k}$. The sample mean $\mu_{i,k}$ and sample variance $\sigma^2_{i,k}$ are our estimates for $\mu_{i, k, true}$ and $\sigma^2_{i,k, true}$, and are computed as follows:

$$
\mu_{i,k} = \frac{1}{M} \sum_{m=1}^M f_{i,k}^m 
$$

$$
\sigma_{i,k} = \frac{1}{M} \sum_{m=1}^M \left(f_{ik}^m - \mu_{ik} \right)^2 = \frac{1}{M} \sum_{m=1}^M f_{ik}^m - \mu_{i,k}^2
$$

Note that $\sigma_{i,k}$ is the **biased** sample variance. Since all we care about is the magnitude of this variance in the end, it doesn't really matter if it's biased or not imo. 

Since the force $k$ force components are independent and normally distributed, we can easily compute the sum of the force components on atom $i$, $f_i = f_{ix} + f_{iy} + f_{iz}$ (since a sum of Gaussians is nice to work with). Then, $f_i$ is also a normally distributed random variable, with (sample) variance $\sigma_{i}^2 = \sigma_{ix}^2 + \sigma_{iy}^2+ \sigma_{iz}^2$. Finally, we define our uncertainty estimate as the atom that has the largest observed sample standard deviation $\sigma_i$, 

$$ \sigma_f = max_i \left( \sqrt{\sigma_i^2} \right) = max_i \left( \sqrt{\sigma_{ix}^2 + \sigma_{iy}^2+ \sigma_{iz}^2} \right) $$

There thus need to have two GPU vectors (since forces are stored on the GPU), of size `3*N` where `N` is the number of atoms, to store $\mu_{i,k}$ and $\sigma_{i,k}$. A third GPU vector of length `N` is also needed to hold $\sigma_i $, which then can be moved to the CPU and the maximum be computed. 

### User API
1. New keyword `active`, behaving similar to `dump_observer`.
2. Supply a list of potentials like for `dump_observer`
3. Second argument to `dump_observer` is the evaluation frequency `l`.
4. Third argument is `has velocity`
5. Fourth argument is `has force`
6. Fifth argument to `dump_observer` is the threshold uncertainty `\delta`.

### Tasks
- [x] Add a new keyword called `active`, using the same logic as `dump_observer`.
- [x] Add `delta` keyword to `active`.
- [x] Add `dump` keyword to `active`.
- [x] Implement two file handlers for `active`, one corresponding to `active.out` and one to `active.xyz`. Copy logic for `active.xyz` from `dump_exyz`.
- [x] Set `multiple_potentials_mode_` to `observe` in `force.cu`; evaluate all potentials manually in `active.cu` every `l` time steps.
- [x] If uncertainty estimate exceeds threshold, write the structure with the **main NEP potential**, `potential[0]` to `active.xyz` in the `exyz` format.
- [x] Write uncertainty to structure in `active.xyz` as well (edit line2).
- [x] Write tests
- [x] Implement new uncertainty model
- [x] Write documentation
- [x] Check if `dump_observer` run in `average` mode will interfere with `active`.


elindgren:
ping @erhart1

elindgren:
Code runs, but there seems to be an issue with calculating the uncertainty. If I read the forces $f_{ik}$ directly from `.xyz`-files written by `dump_observer` and calculate the uncertainty that way (both using Numpy and a Python implementation of my C++ algorithm) I'm off by almost a factor of 2. I'll have to investigate this further.

uncertainty

brucefan1983:
Fantastic work! So we will have on-the-fly active learining! I will study the code.

Dankomaister:
This is great!
We have used active learning with DP-GEN and DeePMD to develop machine learning potentials in our group and this is one of the features we wanted for GPUMD 😍
I see that the mean of the force deviations is currently used to select candidates, are there any plans to also allow selection based on the maximum force deviations, such as what is used in DP-GEN/DeePMD ?

Anyway this is great progress!

brucefan1983:
> Code runs, but there seems to be an issue with calculating the uncertainty. If I read the forces fik directly from `.xyz`-files written by `dump_observer` and calculate the uncertainty that way (both using Numpy and a Python implementation of my C++ algorithm) I'm off by almost a factor of 2. I'll have to investigate this further.
> 
> uncertainty

Perhaps it is because the CUDA kernel tries to use `3*N` threads but is called with `N` threads. So 2/3 of the data were not computed.

brucefan1983:
@elindgren  Can you see my code reviews this time?

elindgren:
> @elindgren Can you see my code reviews this time?

No, not this time either :/ I don't know why the reason for that is. Can you summarize it in regular messages? Or send a picture?

elindgren:
> This is great! We have used active learning with DP-GEN and DeePMD to develop machine learning potentials in our group and this is one of the features we wanted for GPUMD heart_eyes
I see that the mean of the force deviations is currently used to select candidates, are there any plans to also allow selection based on the maximum force deviations, such as what is used in DP-GEN/DeePMD ?

I've thought some more on this, and I believe that using max-force uncertainty instead of the average is actually better outright. I will switch to using max-force, and will thus not implement an extra keyword.

elindgren:
> Perhaps it is because the CUDA kernel tries to use `3*N` threads but is called with `N` threads. So 2/3 of the data were not computed.

That was the issue! It works now

uncertainty

brucefan1983:
> > This is great! We have used active learning with DP-GEN and DeePMD to develop machine learning potentials in our group and this is one of the features we wanted for GPUMD heart_eyes
> > I see that the mean of the force deviations is currently used to select candidates, are there any plans to also allow selection based on the maximum force deviations, such as what is used in DP-GEN/DeePMD ?
> 
> I've thought some more on this, and I believe that using max-force uncertainty instead of the average is actually better outright. I will switch to using max-force, and will thus not implement an extra keyword.

I agree that max is better than mean. The purpose is to detect some  possible large errors.

brucefan1983:
This github feature is really annoying. I don't know why you cannot see my comment. 
image
image
image
image
image
image
image
image
image

elindgren:
> image 

Which forward declaration are you referring to here? I thought I needed to declare `Atom` and `Box` here to be able to use them as types in the other function declarations in `active.cuh`.

Dankomaister:
> > This is great! We have used active learning with DP-GEN and DeePMD to develop machine learning potentials in our group and this is one of the features we wanted for GPUMD heart_eyes
> > I see that the mean of the force deviations is currently used to select candidates, are there any plans to also allow selection based on the maximum force deviations, such as what is used in DP-GEN/DeePMD ?
> 
> @Dankomaister I've thought some more on this, and I believe that using max-force uncertainty instead of the average is actually better outright. I will switch to using max-force, and will thus not implement an extra keyword.

I agree, we have also seen that max force deviation is in practice more suitable for active learning at least with DeePMD.

One thing to note is that you don't want to include unreasonable structures in the active learning process, such as structures that have "exploded" or those were atoms that have collapsed into each other which can happen in the early stages of the active learning process when the MLFF is not that accurate.

DP-GEN does this by having an upper limit on the maximum force deviation that must also be met for the structure to be selected. I am not saying that this is the only way to ensure that unreasonable structures are not selected, one could also, for example, check the minimum interatomic distances. However, some way of detecting unreasonable structures is probably required for the active learning process to be as automated as possible.

brucefan1983:
> > > This is great! We have used active learning with DP-GEN and DeePMD to develop machine learning potentials in our group and this is one of the features we wanted for GPUMD heart_eyes
> > > I see that the mean of the force deviations is currently used to select candidates, are there any plans to also allow selection based on the maximum force deviations, such as what is used in DP-GEN/DeePMD ?
> > 
> > 
> > @Dankomaister I've thought some more on this, and I believe that using max-force uncertainty instead of the average is actually better outright. I will switch to using max-force, and will thus not implement an extra keyword.
> 
> I agree, we have also seen that max force deviation is in practice more suitable for active learning at least with DeePMD.
> 
> One thing to note is that you don't want to include unreasonable structures in the active learning process, such as structures that have "exploded" or those were atoms that have collapsed into each other which can happen in the early stages of the active learning process when the MLFF is not that accurate.
> 
> DP-GEN does this by having an upper limit on the maximum force deviation that must also be met for the structure to be selected. I am not saying that this is the only way to ensure that unreasonable structures are not selected, one could also, for example, check the minimum interatomic distances. However, some way of detecting unreasonable structures is probably required for the active learning process to be as automated as possible.

I  think there is no need to set an upper limit. As long as GPUMD does not crash, all the selected configurations should be taken as potential candidates and should be examined later on. Although the genereal active-learning scheme here is similar to that in DeepMD/D-GEN, I would say we have rather different philosophy in NEP: I think active-learning is only a means to detect possibly missing configurations from an initial dataset that is already quite "complete". Creating a good initial training dataset is still an "art" that one has to master. The NEP models (actually the main one to drive the MD here) should already quite robust. It should be able to run all the MD simulations at all the intended thermodynamic states/process. Also, we should only do the active-learining process by 1-3 times instead of tens of times as common for DeepMD/D-GEN. The final outcome is that, we should keep the number of structures for a given chemical composition (elements or compunds) under, say, 1000.

brucefan1983:
> > image
> 
> Which forward declaration are you referring to here? I thought I needed to declare `Atom` and `Box` here to be able to use them as types in the other function declarations in `active.cuh`.

I meana here if you declare Atom and Box it means you don't need to include the headers `model/atom.cuh` and `model/box.cuh` here. Instead they can be moved to the `active.cu` file.

elindgren:
@brucefan1983 I've added documentation and fixed your comments, so if you think this looks good we can merge :)

shdchen:
Nice work!

PR #398, opened 2023-03-23
Reformatted output for mini-batch predictions

tamaswells:
1.Reformatted output for mini-batch predictions
2.Remove the need of dipole/pol data in prediction mode

### DIFF
--- a/src/main_nep/fitness.cu
+++ b/src/main_nep/fitness.cu
@@ -165,15 +165,18 @@ void Fitness::compute(
   }
 }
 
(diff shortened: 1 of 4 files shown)


brucefan1983:
image

I don't think the change above is needed.

brucefan1983:
I think the method in this PR is too complex. The goal is simple: you want to have an ordered output for dipole and polarizability when predicted using mini-batch. I would suggest a much simpler way to achieve this goal with minimal change to the code.

brucefan1983:
I created a PR here: #399

PR #401, opened 2023-03-26
Allow for the coexistence of stress and virial in the mode of dipole/polarizability predictions

tamaswells:


brucefan1983:
Thanks for the fix.

PR #403, opened 2023-04-02
Add one line output that confirms when the tabulated radial functions are being

hityingph:
To add a line of output that confirms when the tabulated radial functions are being used to speed up NEP.


hityingph:
> To add a line of output that confirms when the tabulated radial functions are being used to speed up NEP.

Upon testing the Carbon example and compiling the gpumd with the -DUSE_TABLE option, I obtained the following output:
image

brucefan1983:
Nice changes.

brucefan1983 (approved):

PR #409, opened 2023-04-18
PIMD-RPMD-TRPMD

brucefan1983:
Usage:

```
ensemble    pimd number_of_beads T_begin T_finish tau_T
ensemble    rpmd number_of_beads 
ensemble    trpmd number_of_beads 
```

### DIFF
--- a/src/integrate/ensemble_pimd.cu
+++ b/src/integrate/ensemble_pimd.cu
@@ -29,20 +29,31 @@ References for implementation:
 #include <cstdlib>
 
 Ensemble_PIMD::Ensemble_PIMD(
+  int number_of_atoms_input, int number_of_beads_input, bool thermostat_internal_input, Atom& atom)
+{
+  number_of_atoms = number_of_atoms_input;
+  number_of_beads = number_of_beads_input;
+  thermostat_internal = thermostat_internal_input;
+  thermostat_centroid = false;
+  initialize(atom);
+}
+
+Ensemble_PIMD::Ensemble_PIMD(
   int number_of_atoms_input,
   int number_of_beads_input,
-  int number_of_steps_pimd_input,
-  double temperature_input,
   double temperature_coupling_input,
   Atom& atom)
 {
   number_of_atoms = number_of_atoms_input;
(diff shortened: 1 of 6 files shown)


src/integrate/ensemble_pimd.cu
```diff
@@ -69,22 +69,33 @@ void Ensemble_PIMD::initialize(Atom& atom)
...
+      atom.position_beads[k].resize(number_of_atoms * 3);
+      atom.velocity_beads[k].resize(number_of_atoms * 3);
+      atom.potential_beads[k].resize(number_of_atoms);
+      atom.force_beads[k].resize(number_of_atoms * 3);
+      atom.virial_beads[k].resize(number_of_atoms * 9);
+
+      atom.position_beads[k].copy_from_device(atom.position_per_atom.data());
+      atom.velocity_beads[k].copy_from_device(atom.velocity_per_atom.data());
+      atom.force_beads[k].copy_from_device(atom.force_per_atom.data());
+    }
```
elindgren, line 98: Why does this need to be wrapped in an if-statement? If `number_of_beads` is 0, would this for loop even execute (since the for-statement is falsy, 0 < 0)?

brucefan1983: > Why does this need to be wrapped in an if-statement? If `number_of_beads` is 0, would this for loop even execute (since the for-statement is falsy, 0 < 0)?
> 
> Here `atom.number_of_beads` and `number_of_beads` are two independent variables. `number_of_beads` must be nonzero up to this point and `atom.number_of_beads` will be zero for the first PIMD run and it will only be equal to `number_of_beads` at line 107 below.

I "edited" again :-)

brucefan1983: The point is that these atom.xxx_beads vectors only need to be allocated and initialized once for the first PIMD run. For later PIMD-related runs, they should not be re-initialized.

elindgren: Ah I see, thanks for the clarification!

PR #413, opened 2023-04-23
start to consider 2 hidden layers

brucefan1983:
Try two hidden layers again to see if it can lead higher accuracy for difficult systems.


brucefan1983:
No gain. So give up (again).

PR #422, opened 2023-05-13
Dump beads

brucefan1983:
## Usage: 
* write this within a PIMD-related run:
```
dump_beads <interval> <has_velocity> <has_force>
```

## Output
* will created `number_of_beads` files named `beads_dump_k.xyz`, where `k` goes from `0` to `number_of_beads - 1`.


shdchen (approved):
Looks good.

PR #424, opened 2023-05-14
Added the `split_xyz.py` script.

initqp:
This script may be invoked to randomly select frames from an existing EXYZ file.


brucefan1983:
Nice work, thanks.

PR #429, opened 2023-05-18
Update singleFrame-outcars2nep-exyz.sh output style

DragonPara:
let ouput 
from:
1/111 2/111 3/111 ... 111/111 dos2unix: converting file to:
as like a loading processing
1/111, and then replace it to 2/111 3/111


Kick-H:
very nice modify, maybe you could also modify the "multi" file

brucefan1983:
Thanks for the contribution @DragonPara and your confimation @Kick-H

PR #441, opened 2023-05-21
Update the tutorials for the predictions of dipole and polarizability.

tamaswells:
Update the tutorials for the predictions of dipole and polarizability.


brucefan1983:
This is superb! Thanks!

PR #442, opened 2023-05-21
add a script for perturbation structures

Baristali:
a simple srcipt for generating strucures with random perturbation in lattice.


Kick-H:
Thanks for this, I'll test it right away

PR #443, opened 2023-05-21
Add cp2k2xyz.py to tools

Kick-H:
Submit Mr. Chen's CP2K2xyz file to the master, and make a simple modification to make it more convenient to use.


brucefan1983:
nice, thanks!

PR #449, opened 2023-05-23
write multicomponent system hnemd in doc

liangzhixin-202169:


brucefan1983:
This is very good. Perhaps you cannot see my online comments. Here is one:
image

PR #453, opened 2023-05-23
only one title in a jupyter notebook

brucefan1983:


brucefan1983:
so I think this can be abandoned.

PR #458, opened 2023-05-26
use model_type

brucefan1983:
* Change the `mode` keyword in `nep.in` to the better name `model_type`. 

@erhart1 I have still kept to accept `mode` for `model_type` to not break some user's inputs. But the documentation has been changed. Users will gradually change I think. I discussed with others and they think `prediciton` does not need to be changed as there is no confusion with `mode` any more. 

My changes are complete and are ready for review.


tamaswells (approved):

erhart1:
Good changes. I am not as fast as you ;-)

PR #461, opened 2023-06-04
compute_rdf_yong_1

AmbroseWong:


src/measure/rdf.cu
```diff
@@ -0,0 +1,607 @@
...
+#include "model/atom.cuh"
+
+
+namespace
+{
+
+static __global__ void gpu_find_rdf_ON1(
+  const int N,
+  const double density,
+  const Box& box,
```
brucefan1983, line 38: Cannot pass reference to kernel. This must trigger illegal memory access.

brucefan1983 (changes requested):

AmbroseWong:
image

I have tested ```compute_rdf``` command and have compared its results with those calculated from OVITO code (see above FIG). The system I simulated is 300 Kelvin water both using classical MD and 32 beads PIMD.

The explicit ```compute_rdf``` command format is:

```compute_rdf r_rdf_cutoff N_bins Num_last_steps Num_repeat Num_interval (atom idx1 idx2 atom idx3 idx4... )```

where ```r_rdf_cutoff``` is cutoff_radius for rdf calculations, ```N_bins``` is how many slices the r_rdf_cutoff will be cut, ```Num_last_steps``` defines the number of last steps range used to calculated rdf, ```Num_repeat``` is the number of trajectory frames and  ```Num_interval``` represents the interval step we chose one frame from ```Num_last_steps``` frames. It should be noted that ```Num_last_steps = Num_repeat * Num_interval``` and  ```Num_last_steps``` can not be larger than the number of total simulation steps. 

Using the above parameters, the total rdf including all elements will be calculated and saved in ***rdf_total.out*** or ***rdf_centroid_total.out*** if path integral schemes were used. 

By adding optional parameters with the group unit ```atom idx1 idx2```,  the element-dependent rdf will be calculated. Elemental indexes used in this command (***indx1...***) are the same as definitions in the ***potential file*** (like ***nep.txt***). For instance, supposing ***0*** represents hydrogen element ***H*** and ***1*** represents oxygen element ***O***, if we use ```atom 0 1``` here, then an extra file  ***rdf_type_0_1.out*** or ***rdf_centroid_type_0_1.out*** will be calculated and written, which will store the ***H-O*** rdf. At present, **the maximum number** of ```atom idx1 idx2``` is set to **6** to avoid too many memory occupations.

AmbroseWong:
image

I have tested ```compute_rdf``` command and have compared its results with those calculated from OVITO code (see above FIG). The system I simulated is 300 Kelvin water both using classical MD and 32 beads PIMD.

The explicit ```compute_rdf``` command format is:

```compute_rdf r_rdf_cutoff N_bins Num_interval (atom idx1 idx2 atom idx3 idx4... )```

where ```r_rdf_cutoff``` is cutoff_radius for rdf calculations, ```N_bins``` is how many slices the r_rdf_cutoff will be cut,  and ```Num_interval``` represents the interval step with which we chose frames from equilibrium process to calculate ensemble average. 

Using the above essential parameters, the total rdf including all elements will be calculated and saved in ***rdf.out*** or ***rdf_centroid.out*** if path integral schemes were used. See data details in the first row of this file.

By adding optional parameters with the command unit ```atom idx1 idx2```,  the element-dependent rdf will be calculated. Elemental indexes used in this command (***indx1...***) are the same as definitions in the ***potential file*** (like ***nep.txt***). For instance, supposing ***0*** represents hydrogen element ***H*** and ***1*** represents oxygen element ***O***, if we use ```atom 0 1``` here, then an extra column  ***type_0_1*** will be calculated and written into the ***RDF output file***, which will store the ***H-O*** rdf. 

At present, **the maximum set number** of ```atom idx1 idx2``` is set to **6** to avoid too many memory occupations.

For large box systems, a large ```Num_interval``` is recommended to save computation time and the size of systems has already guaranteed a good statistical average.

brucefan1983 (approved):

PR #471, opened 2023-07-30
add dipole and polarizability info to neptxt

brucefan1983:
Add dipole and polarizability info to `nep.txt` such that it can be read later by `gpumd` or `NEP_CPU`.


erhart1:
Looks good!
This should make parsing much simpler.

PR #474, opened 2023-08-18
electron_entropy

XIX-YANG:
add electron temperature in train and md


brucefan1983:
Need to start from the master version. I will close this and we do it in a new PR.

PR #478, opened 2023-08-25
add option to dump potential

brucefan1983:
Allowed syntax:

```
dump_xyz <interval> # only position

# should be clear below:
dump_xyz <interval>  has_velocity 
dump_xyz <interval>  has_velocity has_force
dump_xyz <interval>  has_velocity has_force has_potential
```

Still not very beautiful but ok for the time being. I aimed to be backward compatible.

### DIFF
--- a/src/measure/dump_exyz.cu
+++ b/src/measure/dump_exyz.cu
@@ -18,6 +18,7 @@ Dump some data to dump.xyz in the extended XYZ format
 --------------------------------------------------------------------------------------------------*/
 
 #include "dump_exyz.cuh"
+#include "model/atom.cuh"
 #include "model/box.cuh"
 #include "utilities/common.cuh"
 #include "utilities/error.cuh"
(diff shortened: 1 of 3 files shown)


erhart1:
Nice. I had missed this. I will test it for our coexistence runs

brucefan1983:
I forgot to add something in line 2, just created as PR: https://github.com/brucefan1983/GPUMD/pull/492/files

But what should be the key?  potential? energy?

erhart1:
`energy_atom` works. `energy_per_atom` might be even clearer.

erhart1:
Would it be possible to add support for the `dump_position` command as well?

erhart1:
It appears that the `energy_per_atom` is not included in the `Properties` field of the extended xyz format line.
I get `Properties=species:S:1:pos:R:3` but the lines below do include the potential energy per atom.

brucefan1983:
Are you using a version before this
commit: 4ccac3a362fb47ae11827a78816526e0dfc94a41

Zheyong

brucefan1983:
> Would it be possible to add support for the `dump_position` command as well?

perhaps leave it to the breaking version 4.0, where we can suport multiple dump_exyz?

erhart1:
> Are you using a version before this commit: https://github.com/brucefan1983/GPUMD/commit/4ccac3a362fb47ae11827a78816526e0dfc94a41

Yes. I was actually running from the `sgc` branch. That probably explains.

brucefan1983:
> > Are you using a version before this commit: [4ccac3a](https://github.com/brucefan1983/GPUMD/commit/4ccac3a362fb47ae11827a78816526e0dfc94a41)
> 
> Yes. I was actually running from the `sgc` branch. That probably explains.

It was added to master. So you will get it later.

PR #518, opened 2023-10-25
Non-equilibrium free energy integration

psn417:
I have compared the result of gpumd and lammps with a short simulation:

gpumd:
image
lammps:
image

gpumd dW = -7.6753184549175 eV
lammps dW = -7.675099831441699 eV


brucefan1983:
This is nice work. Thanks!

PR #523, opened 2023-11-02
minor bug fix

psn417:
add fix atoms to mttk ensemble

### DIFF
--- a/doc/gpumd/input_parameters/velocity.rst
+++ b/doc/gpumd/input_parameters/velocity.rst
@@ -11,9 +11,10 @@ Syntax
 ------
 * This keyword only has one parameter, which is the initial temperature of the system::
   
-    velocity <initial_temperature>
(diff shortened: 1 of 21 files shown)


psn417:
The line of code I added seems to be duplicate with the codes elsewhere. Perhaps I'd better delete them all.

psn417:
# Summary
In this PR, I've done the following things:
- Add fix atoms for MTTK thermostat.
- The `fixed_group` is passed to the `ensemble` base class directly, so there is no need to pass it to each constructor of subclasses. I deleted the corresponding arguments in the constructors.
- The group information was not handled correctly in `replicate` command. I fixed it.
- I added an optional `seed` keyword to `velocity` command to produce identical velocities across different runs. It works correctly. 
After modifying the code, I test the fix command in `nve`, `nvt_nhc`, and `mttk` command, together with the `replicate` command. The atoms are fixed correctly, as shown below:
image

PR #530, opened 2023-11-26
add a function to make the compute_shc command can compute all group in one run.

Kick-H:


Kick-H:
shc_multi
I have test the command with two version.
I just modify the two file named measure/shc.cu and measure/shc.cuh.

PR #533, opened 2023-12-11
Dump dipole & polarizability during md

### LINKED ISSUE #407: Predict dipole during MD
It would be very useful to be able to compute the dipole during MD. My idea was to use `dump_observer`, where one would supply two models, `nep_main` and `nep_dipole`. `nep_main` would be used to run the MD and `nep_dipole` would be used to compute the dipole as an observer model. 

In order to implement the above scheme we would need a function for computing dipoles in, for instance, `force/nep3.cu`. Also, it would be beneficial if there was a flag in the `nep.txt` file that indicated what type of model it was (regular, polarizability, dipole), which I don't believe is done atm? @brucefan1983 Have you been planning to implement a function for calculating dipoles during MD?

elindgren:
Dump dipoles and polarizability during MD, similar to `dump_observer`.
The keywords should be `dump_dipole` and `dump_polarizability`, with one input parameter:
- (int) no of timesteps between dumps.

Will write one file for dipoles and polarizabilities each, `dipole.out` and `polarizability.out`. The format should be as follows:
- dipole.out: 4 columns, with time step, $x$, $y$ and $z$ components.
- polarizability.out: 7 columns, for time step and $xx$, $yy$, $zz$, $xy$, $yz$, $zx$ components. 
for each row. 

One needs to specify two models, where the second will be a dipole or polarizability model.

Checklist
- [x] Add keyword `dump_dipole`
- [x] Add keyword `dump_polarizability`
- [x] Parse the potential file to deduce if a dipole or polarizability model has been supplied, or both. 
- [x] Add functionality to read nep3 dipole models
- [x] Add functionality to read nep4 dipole models
- [x] Make changes to
    - [x] `nep3.cu`
    - [x] `nep3_multigpu.cu`
    - [x] `nep3_small_box.cu`
- [x] Prepare tests for `dump_dipole`
    - [x] Compare to dipole values saved on file
    - [x] Compare with predictions from the `NEP` executable to ensure self-consistency. 
- [x] Prepare tests for `dump_polarizability`
    - [x] Compare to polarizability values saved on file
    - [x] Compare with predictions from the `NEP` executable to ensure self-consistency. 
- [x] Implement equations for dipoles in the GPUMD executable. -> equation 3
- [x] Implement equations for polarizability in the GPUMD executable. -> equation 4, energy - virial.
- [x] Add documentation for `dump_dipole`
- [x] Add documentation for `dump_polarizability`
- [x] Optimize kernels 
- [ ] Figure out why regression test for `dump_observer` is broken (it's failing in master as well)

Fixes #407


elindgren:
Hey @brucefan1983, I've started working on this. I have something that runs and compiles for the dipoles, but the predicted values are a bit off when I compare to `NEP_CPU` via `calorine. Could you take a quick look and see if I've missed/screwed up the implementation somewhere?

I also have a specific question, regarding the `find_force_angular` equations. Shouldn't the virials be computed in all instances of `find_force_angular*` in `nep3.cu`, as they are in `main_nep/nep3.cu::find_force_angular`? As far as I can tell, they are only in `find_force_angular` in `nep3_small_box.cuh`.

brucefan1983:
> Hey @brucefan1983, I've started working on this. I have something that runs and compiles for the dipoles, but the predicted values are a bit off when I compare to `NEP_CPU` via `calorine. Could you take a quick look and see if I've missed/screwed up the implementation somewhere?
> 
> I also have a specific question, regarding the `find_force_angular` equations. Shouldn't the virials be computed in all instances of `find_force_angular*` in `nep3.cu`, as they are in `main_nep/nep3.cu::find_force_angular`? As far as I can tell, they are only in `find_force_angular` in `nep3_small_box.cuh`.

It depends on if `atomicAdd()` is used. In case it is used, virial is calculated within a single kernel function, otherwise there is another function to complete the evaluation. In NEP part and GPUMD small-box part, I used `atomicAdd()`, but in other cases, I have not used `atomicAdd()`. The reason is related to speed and also the preference of no randomness in production MD runs. atomicAdd()` introduces randomness.

elindgren:
> In case it is used, virial is calculated within a single kernel function, otherwise there is another function to complete the evaluation

I see. What is the name of the other function? I haven't been able to find where it.

elindgren:
I've also started working on implementing polarizabilities. Are they just the summed virials with a `_pol` model? It looks like the pol is just the virials, for instance here in `NEP_CPU`:

```cpp
for (int d = 0; d < 6; ++d) {
    polarizability[d] = 0.0;
  }
  for (int n = 0; n < N; ++n) {
    polarizability[0] += virial[0 * N + n]; // xx
    polarizability[1] += virial[4 * N + n]; // yy
    polarizability[2] += virial[8 * N + n]; // zz
    polarizability[3] += virial[1 * N + n]; // xy
    polarizability[4] += virial[5 * N + n]; // yz
    polarizability[5] += virial[6 * N + n]; // zx
  }
```
In other words, do I just have to predict the virials with a polarizability and just sum it up over atoms? If I read the nep dipole paper correctly, I would have expected the potential energy to have been subtracted from the diagonals:

image

I've tried both approaches in `gpumd`, both predicting the virial and the potential - virial, but I don't get the same values as in `NEP_CPU`. Am I missing something?

Here is my implementation for calculating the polarizability:

https://github.com/brucefan1983/GPUMD/blob/46df9a7982ca8449ea97eb15c31dd681b0042981/src/measure/dump_polarizability.cu#L142-L172

brucefan1983:
> > In case it is used, virial is calculated within a single kernel function, otherwise there is another function to complete the evaluation
> 
> I see. What is the name of the other function? I haven't been able to find where it.

It is `void Potential::find_properties_many_body` at line 272 of `force/potential.cu`

brucefan1983:
> @brucefan1983 I've also started working on implementing polarizabilities. Are they just the summed virials with a `_pol` model? It looks like the pol is just the virials, for instance here in `NEP_CPU`:
> 
> ```c++
> for (int d = 0; d < 6; ++d) {
>     polarizability[d] = 0.0;
>   }
>   for (int n = 0; n < N; ++n) {
>     polarizability[0] += virial[0 * N + n]; // xx
>     polarizability[1] += virial[4 * N + n]; // yy
>     polarizability[2] += virial[8 * N + n]; // zz
>     polarizability[3] += virial[1 * N + n]; // xy
>     polarizability[4] += virial[5 * N + n]; // yz
>     polarizability[5] += virial[6 * N + n]; // zx
>   }
> ```
> 
> In other words, do I just have to predict the virials with a polarizability and just sum it up over atoms? If I read the nep dipole paper correctly, I would have expected the potential energy to have been subtracted from the diagonals:
> 
> image
> 
> I've tried both approaches in `gpumd`, both predicting the virial and the potential - virial, but I don't get the same values as in `NEP_CPU`. Am I missing something?
> 
> Here is my implementation for calculating the polarizability:
> 
> https://github.com/brucefan1983/GPUMD/blob/46df9a7982ca8449ea97eb15c31dd681b0042981/src/measure/dump_polarizability.cu#L142-L172
> 
> This is the kernel I use to compute the polarizability from the virials:
> 
> https://github.com/brucefan1983/GPUMD/blob/46df9a7982ca8449ea97eb15c31dd681b0042981/src/measure/dump_polarizability.cu#L30-L51

The paper is right (must be), and in NEP_CPU, the potential part is done in the function `find_descriptor_small_box` (line 847 of `nep.cpp` in NEP_CPU repo). You can check the option when `bool calculating_polarizability` is true. The scaler part should have been done in this function, with a new set of NN parameters (not the same NN parameters for the tensor part).

brucefan1983:
I will have a full-week holiday and will look into the code as soon as I have free time. @elindgren

src/force/nep3.cu
```diff
@@ -123,6 +123,18 @@ NEP3::NEP3(const char* file_potential, const int num_atoms)
...
+  } else if (tokens[0] == "nep3_dipole") {
+    paramb.version = 3;
+    paramb.model_type = 1;
+  } else if (tokens[0] == "nep3_polarizability") {
+    paramb.version = 3;
+    paramb.model_type = 2;
+  } else if (tokens[0] == "nep4_dipole") {
+    paramb.version = 4;
+    paramb.model_type = 1;
+  } else if (tokens[0] == "nep3_polarizability") {
```
brucefan1983, line 135: `nep3_polarizability` should be `nep4_polarizability`

src/force/nep3_multigpu.cu
```diff
@@ -104,6 +104,18 @@ NEP3_MULTIGPU::NEP3_MULTIGPU(
...
+  } else if (tokens[0] == "nep3_dipole") {
+    paramb.version = 3;
+    paramb.model_type = 1;
+  } else if (tokens[0] == "nep3_polarizability") {
+    paramb.version = 3;
+    paramb.model_type = 2;
+  } else if (tokens[0] == "nep4_dipole") {
+    paramb.version = 4;
+    paramb.model_type = 1;
+  } else if (tokens[0] == "nep3_polarizability") {
```
brucefan1983, line 116: `nep3_polarizability` should be `nep4_polarizability`

src/measure/dump_dipole.cu
```diff
@@ -0,0 +1,172 @@
...
+#include "utilities/gpu_vector.cuh"
+#include "utilities/read_file.cuh"
+#include <iostream>
+#include <vector>
+
+static __global__ void sum_diagonal(int N, const double* g_virial_per_atom, double* g_virial_sum)
+{
+  int n1 = blockIdx.x * blockDim.x + threadIdx.x;
+  if (n1 < N) {
+    g_virial_sum[0] += g_virial_per_atom[n1 + 0 * N];
```
brucefan1983, line 34: The summation here is no correct. You need to either use `atomicAdd`, or use the so-called binary reduction. You can check e.g., the function, `gpu_sum_tensor()` in `force.cu`.

src/measure/dump_polarizability.cu
```diff
@@ -0,0 +1,203 @@
...
+#include "parse_utilities.cuh"
+#include "utilities/common.cuh"
+#include "utilities/error.cuh"
+#include "utilities/gpu_vector.cuh"
+#include "utilities/read_file.cuh"
+#include <iostream>
+#include <vector>
+
+static __global__ void sum_polarizability(
+  int N, const double* g_potential_per_atom, const double* g_virial_per_atom, double* g_pol)
```
brucefan1983, line 31: The summation here is no correct. You need to either use `atomicAdd`, or use the so-called binary reduction. You can check e.g., the function, `gpu_sum_tensor()` in `force.cu`.

src/measure/dump_polarizability.cu
```diff
@@ -0,0 +1,203 @@
...
+      cpu_pol_[i] += cpu_pol_per_atom_[j + i * number_of_atoms];
+    }
+  }
+  // Write properties
+  write_polarizability(step);
+}
+
+void Dump_Polarizability::write_polarizability(const int step)
+{
+  if (!dump_)
```
brucefan1983, line 177: no need to check any more. It has been checked in the calling function.

src/measure/dump_dipole.cu
```diff
@@ -0,0 +1,172 @@
...
+
+  // Transfer gpu_sum to the CPU
+  gpu_dipole_.copy_to_host(cpu_dipole_.data());
+  // Write properties
+  write_dipole(step);
+}
+
+void Dump_Dipole::write_dipole(const int step)
+{
+  if (!dump_)
```
brucefan1983, line 156: no need to check any more. It has been checked in the calling function.

elindgren:
@brucefan1983 Thanks for the review and the tips! I think I've gotten the dipoles working now using `atomicAdd` (just a factor of -1 off), but the polarizability still eludes me. I'm working on implementing `find_descriptor*` as in `NEP_CPU`, such that the polarizability weights `w*_pol` are being used. Are the `_pol` weights placed after the regular weights in `nep.txt`? From `update_potential` in `NEP_CPU` it looks like the pointer is being shifted:

```cpp
void NEP3::update_potential(double* parameters, ANN& ann)
{
  double* pointer = parameters;
  for (int t = 0; t < paramb.num_types; ++t) {
    if (t > 0 && paramb.version != 4) { // Use the same set of NN parameters for NEP2 and NEP3
      pointer -= (ann.dim + 2) * ann.num_neurons1;
    }
    ann.w0[t] = pointer;
    pointer += ann.num_neurons1 * ann.dim;
    ann.b0[t] = pointer;
    pointer += ann.num_neurons1;
    ann.w1[t] = pointer;
    pointer += ann.num_neurons1;
  }

  ann.b1 = pointer;
  pointer += 1;

  if (paramb.model_type == 2) {
    for (int t = 0; t < paramb.num_types; ++t) {
      if (t > 0 && paramb.version != 4) { // Use the same set of NN parameters for NEP2 and NEP3
        pointer -= (ann.dim + 2) * ann.num_neurons1;
      }
      ann.w0_pol[t] = pointer;
      pointer += ann.num_neurons1 * ann.dim;
      ann.b0_pol[t] = pointer;
      pointer += ann.num_neurons1;
      ann.w1_pol[t] = pointer;
      pointer += ann.num_neurons1;
    }
    ann.b1_pol = pointer;
    pointer += 1;
  }

  ann.c = pointer;
}
```

PetterRosander:
@brucefan1983 @elindgren will it be possible to use this tag with pimp (for the centroid) or would that require extra development?

elindgren:
@

> @brucefan1983 @elindgren will it be possible to use this tag with pimp (for the centroid) or would that require extra development?

If the PIMD implementation uses the same functions to compute the virial for the centroid as in this function then I expect it to work. I'm not familiar with the PIMD implementation though, so I don't know how the virials are handled.

src/force/nep3.cu
```diff
@@ -794,15 +855,22 @@ static __global__ void find_force_radial(
       s_fx += f12[0] - f21[0];
       s_fy += f12[1] - f21[1];
       s_fz += f12[2] - f21[2];
-      s_sxx += r12[0] * f21[0];
+      if (is_dipole) {
+        double r12_square = r12[0] * r12[0] + r12[1] * r12[1] + r12[2] * r12[2];
+        s_sxx += r12_square * f21[0];
```
brucefan1983, line 860: should be `s_sxx += r12_square * f21[0];` etc, and all other similar places for dipole

src/force/nep3_multigpu.cu
```diff
@@ -913,9 +946,31 @@ static __global__ void find_descriptor(
...
+        annmb.b1_pol,
+        q,
+        F,
+        Fp);
+      // Add the potential values to the diagonal of the virial
+      g_virial[n1] = F;
+      g_virial[n1 + N * 1] = F;
+      g_virial[n1 + N * 2] = F;
+
+      for (int d = 0; d < annmb.dim; ++d) {
```
brucefan1983, line 966: Also set F back to 0.0f

src/force/nep3_multigpu.cu
```diff
@@ -913,9 +946,31 @@ static __global__ void find_descriptor(
...
+        Fp[d] = 0.0;
+      }
+    }
+
     apply_ann_one_layer(
       annmb.dim, annmb.num_neurons1, annmb.w0[t1], annmb.b0[t1], annmb.w1[t1], annmb.b1, q, F, Fp);
-    g_pe[n1] = F;
+    g_pe[n1] += F;
```
brucefan1983, line 973: Seems this does not need to be `+=`. Just `=` is as before.

src/force/nep3_multigpu.cu
```diff
@@ -1044,15 +1100,22 @@ static __global__ void find_force_radial(
       s_fx += f12[0] - f21[0];
       s_fy += f12[1] - f21[1];
       s_fz += f12[2] - f21[2];
-      s_sxx += r12[0] * f21[0];
+      if (is_dipole) {
+        float r12_square = r12[0] * r12[0] + r12[1] * r12[1] + r12[2] * r12[2];
+        s_sxx += r12_square * f21[0];
```
brucefan1983, line 1105: `+=` to `-=` as commented elsewhere

src/force/nep3_multigpu.cu
```diff
@@ -2009,6 +2079,28 @@ static __global__ void find_descriptor(
...
+        annmb.w1_pol[t1],
+        annmb.b1_pol,
+        q,
+        F,
+        Fp);
+      // Add the potential values to the diagonal of the virial
+      g_virial[n1] = F;
+      g_virial[n1 + N * 1] = F;
+      g_virial[n1 + N * 2] = F;
+
```
brucefan1983, line 2098: set F back to 0.0f;

src/force/nep3_small_box.cuh
```diff
@@ -287,6 +289,28 @@ static __global__ void find_descriptor_small_box(
...
+        annmb.w1_pol[t1],
+        annmb.b1_pol,
+        q,
+        F,
+        Fp);
+      // Add the potential values to the diagonal of the virial
+      g_virial[n1] = F;
+      g_virial[n1 + N * 1] = F;
+      g_virial[n1 + N * 2] = F;
+
```
brucefan1983, line 308: set F back to 0.0f;

src/force/nep3_small_box.cuh
```diff
@@ -441,6 +467,28 @@ static __global__ void find_descriptor_small_box(
...
```
brucefan1983, line 486: set F back to 0.0f;

src/force/potential.cu
```diff
@@ -237,15 +244,23 @@ static __global__ void gpu_find_force_many_body(
       s_fz += f12z - f21z;
 
       // per-atom virial
-      s_sxx += x12 * f21x;
+      if (is_dipole) {
+        // Float version of the function
+        float r12_square = x12 * x12 + y12 * y12 + z12 * z12;
+        s_sxx += r12_square * f21x;
```
brucefan1983, line 250: `+=` to `-=` as commented elsewhere

brucefan1983:
For dipole, there is a global sign error as I commented for the code.

For polarizability, the problem is that the extra set of NN parameters has not been counted to the total number of parameters. Please check the following lines in NEP_CPU repo:

image

Therefore, the code is actually dangerous, trying to accessing non-existing parameter data. It did not crash just because of luck.

With the obove fixes, you should make the code working for both dipole and polarizability. We will discuss about computational performance later.

brucefan1983:
Perfect. You might want to do more tests on polarizability, as the example you provided has almost zero off-diagonal elements. 

I will comment on some speed aspects later.

elindgren:
> Perfect. You might want to do more tests on polarizability, as the example you provided has almost zero off-diagonal elements.
> 
> I will comment on some speed aspects later.

Ah yes you're right. Maybe @PetterRosander has a suitable example system I could use?

src/measure/dump_dipole.cu
```diff
@@ -0,0 +1,171 @@
...
+#include "utilities/common.cuh"
+#include "utilities/error.cuh"
+#include "utilities/gpu_vector.cuh"
+#include "utilities/read_file.cuh"
+#include <iostream>
+#include <vector>
+
+static __global__ void sum_diagonal(int N, const double* g_virial_per_atom, double* g_virial_sum)
+{
+  // TODO benchmark atomics, consider reductions
```
brucefan1983, line 32: Yes, reduction should be faster.

brucefan1983: you can use <<<3, 1024>>>

src/measure/dump_dipole.cu
```diff
@@ -0,0 +1,171 @@
...
+    g_virial[n1 + 1 * N] = 0.0;
+    g_virial[n1 + 2 * N] = 0.0;
+    g_virial[n1 + 3 * N] = 0.0;
+    g_virial[n1 + 4 * N] = 0.0;
+    g_virial[n1 + 5 * N] = 0.0;
+    g_virial[n1 + 6 * N] = 0.0;
+    g_virial[n1 + 7 * N] = 0.0;
+    g_virial[n1 + 8 * N] = 0.0;
+  }
+  g_virial_sum[0] = 0.0;
```
brucefan1983, line 66: better to confine this to a single thread

src/measure/dump_polarizability.cu
```diff
@@ -0,0 +1,199 @@
...
+    int z = n1 + 2 * N;
+    int xy = n1 + 3 * N;
+    int yz = n1 + 5 * N;
+    int zx = n1 + 7 * N;
+    // Sum up per atom polarizability
+    // xx yy zz xy yz zx
+    // atomicAdd(&g_pol[n1 + 0 * N], g_potential_per_atom[x] - g_virial_per_atom[x]); // xx
+    // atomicAdd(&g_pol[n1 + 1 * N], g_potential_per_atom[y] - g_virial_per_atom[y]); // yy
+    // atomicAdd(&g_pol[n1 + 2 * N], g_potential_per_atom[z] - g_virial_per_atom[z]); // zz
+    atomicAdd(&g_pol[0], g_virial_per_atom[x]);  // xx
```
brucefan1983, line 47: try to use reduction with `<<<6, 1024>>>`

elindgren:
@brucefan1983 I've implemented array reductions for summing in `dump_dipole` and `dump_polarizability`. However, when benchmarking it seems like this part of the code has negligible contribution to overall execution time for a relatively large system (~550K atoms) when I dump dipole/pol every iteration. See the table below.

| Case      | Time(s) |
| ----------- | ----------- |
| Baseline (no dump_pol keyword)     | 51 |
| Calculate but don't write pol             | 69 |
| Calculate force only                         | 69 |
| No force calculation                         | 50 |
| AtomicAdd                                       | 69 |
| Reduction                                        | 71 |

Note that the values in the table aren't averaged over multiple runs, so I would expect a variability of a few second between different runs.  As you can see it is the `force` evaluation that takes almost all the extra time. I implemented an array reduction anyways, but I don't see any point in optimizing it further.

PetterRosander:
> @brucefan1983 @elindgren will it be possible to use this tag with pimp (for the centroid) or would that require extra development?

ping @brucefan1983

Without any expectations, I've tested this with PIMD, and it gave me nonsense predictions.
From a UI perspective, I think this tag should be useable with all ensembles, is this possible to fix?

brucefan1983:
> > @brucefan1983 @elindgren will it be possible to use this tag with pimp (for the centroid) or would that require extra development?
> 
> ping @brucefan1983
> 
> Without any expectations, I've tested this with PIMD, and it gave me nonsense predictions. From a UI perspective, I think this tag should be useable with all ensembles, is this possible to fix?

Could you send me a complete example that I can run and reproduce (only the GPUMD part, as I do not use python)?

brucefan1983:
@elindgren do you want to resolve the conflicts with master now?

PetterRosander:
> > > @brucefan1983 @elindgren will it be possible to use this tag with pimp (for the centroid) or would that require extra development?
> > 
> > 
> > ping @brucefan1983
> > Without any expectations, I've tested this with PIMD, and it gave me nonsense predictions. From a UI perspective, I think this tag should be useable with all ensembles, is this possible to fix?
> 
> Could you send me a complete example that I can run and reproduce (only the GPUMD part, as I do not use python)?

There is no python involved, this is only using gpumd with this new implementation.
Ill attach the run.in that I've used
[run.in](https://github.com/brucefan1983/GPUMD/files/14113881/run.md)

brucefan1983:
> onsense predictio

By `nonsense` what do you mean exactly? You mean the predict polarizability then gives nonsense Raman? If so, it might be due to the fact you were running with PIMD, not RPMD or TRPMD.

What you can test is the consistency between the following two sets of data:

* predicted nonsense values directly from the dump command
* predicted nonsense values based on the trajectory produced by `dump_exyz`, which contains the centroid degree of freedom in PIMD-related runs.

PetterRosander:
> > onsense predictio
> 
> By `nonsense` what do you mean exactly? You mean the predicted polarizability then gives nonsense Raman? If so, it might be due to the fact you were running with PIMD, not RPMD or TRPMD.
> 
> What you can test is the consistency between the following two sets of data:
> 
> * predicted polarizability  values directly from the dump command
> * predicted polarizability  values based on the trajectory produced by `dump_exyz`, which contains the centroid degree of freedom in PIMD-related runs.
> 
> Moreover, it should be noted that PIMD-related runs also require a more "robust" NEP potential model than usual, as the quantum motions are usually more violent than the classical ones. If you trained a NEP potential model based on calssical configurations, it does not garantee a very robust PIMD-related simulation. Have you checked the trajectories of the beads?

By nonsense, I just mean the predicted value of the polarizability, it gives results several order of magnitudes of what is expected, i.e., junk values.
However, now that tried it again it seems to give more reasonable values, not sure what I did wrong the first time.
But just to clarify, as the code is written now, this gives me the polarizabilities of the centroid, right?

brucefan1983:
> > > onsense predictio
> > 
> > 
> > By `nonsense` what do you mean exactly? You mean the predicted polarizability then gives nonsense Raman? If so, it might be due to the fact you were running with PIMD, not RPMD or TRPMD.
> > What you can test is the consistency between the following two sets of data:
> > 
> > * predicted polarizability  values directly from the dump command
> > * predicted polarizability  values based on the trajectory produced by `dump_exyz`, which contains the centroid degree of freedom in PIMD-related runs.
> > 
> > Moreover, it should be noted that PIMD-related runs also require a more "robust" NEP potential model than usual, as the quantum motions are usually more violent than the classical ones. If you trained a NEP potential model based on calssical configurations, it does not garantee a very robust PIMD-related simulation. Have you checked the trajectories of the beads?
> 
> By nonsense, I just meant the predicted value of the polarizability, it gave me results several order of magnitudes larger than what I expected, i.e., junk values. However, now that tried it again it seems to give more reasonable values, not sure what I did wrong the first time. But just to clarify, as the code is written now, this gives me the polarizabilities of the centroid, right?

yes. and if you want to do correlation function, you need PRMD/TRPMD.

elindgren:
> @elindgren do you want to resolve the conflicts with master now?

Yes will do! I'm currently trying to fix a problem with one of the regression tests I wrote for `dump_observer`, will hopefully have both fixed soon. Then I just need to add the docs and we can merge.

elindgren:
@brucefan1983 I've added documentation for this and I think we can merge this if you're happy.

Unfortunately I haven't been able to figure out the source of the failing integration test for dump_observer:average, but it is equally broken on master. I suspect something weird happens when I evaluate multiple potentials after each other, but I think we should continue to investigate this in a separate pr.

src/force/potential.cu
```diff
@@ -100,14 +100,20 @@ static __global__ void gpu_find_force_many_body(
 
       // per-atom virial
       s_sxx += x12 * f21x;
+      s_syy += y12 * f21y;
```
brucefan1983, line 103: This function is for non-NEP potentials, and should  be brought back to its original form.

brucefan1983: Here some virial components are double accumulated.

src/force/force.cu
```diff
@@ -92,6 +92,10 @@ void Force::parse_potential(
     strcmp(potential_name, "nep") == 0 || strcmp(potential_name, "nep_zbl") == 0 ||
     strcmp(potential_name, "nep3") == 0 || strcmp(potential_name, "nep3_zbl") == 0 ||
     strcmp(potential_name, "nep4") == 0 || strcmp(potential_name, "nep4_zbl") == 0 ||
+    strcmp(potential_name, "nep3_dipole") == 0 ||
+    strcmp(potential_name, "nep3_polarizability") == 0 ||
+    strcmp(potential_name, "nep4_dipole") == 0 ||
+    strcmp(potential_name, "nep4_polarizability") == 0 ||
```
brucefan1983, line 98: Also consider `nep_dipole` and `nep_polarizability`, which are based on "version 2" of NEP.

src/force/nep3_small_box.cuh
```diff
@@ -287,6 +289,29 @@ static __global__ void find_descriptor_small_box(
...
+      F = 0.0;
```
brucefan1983, line 309: Use `0.0f` to avoid type conversion, which can save time.

src/force/nep3_small_box.cuh
```diff
@@ -287,6 +289,29 @@ static __global__ void find_descriptor_small_box(
...
+      for (int d = 0; d < annmb.dim; ++d) {
+        Fp[d] = 0.0;
```
brucefan1983, line 311: Use 0.0f to avoid type conversion, which can save time.

src/force/nep3_small_box.cuh
```diff
@@ -315,12 +340,14 @@ static __global__ void find_descriptor_small_box(
   const float* __restrict__ g_x12_angular,
   const float* __restrict__ g_y12_angular,
   const float* __restrict__ g_z12_angular,
+  const bool is_polarizability,
```
brucefan1983, line 343: There is no need to modify the functions for the "temperature" version because in this case `nep_model_type=3`.

src/force/nep3.cu
```diff
@@ -1482,12 +1566,14 @@ static __global__ void find_descriptor(
   const double* __restrict__ g_x,
   const double* __restrict__ g_y,
   const double* __restrict__ g_z,
+  const bool is_polarizability,
```
brucefan1983, line 1569: There is no need to modify the functions for the "temperature" version because in this case `nep_model_type=3`.

src/measure/dump_dipole.cu
```diff
@@ -0,0 +1,237 @@
...
+#include "model/read_xyz.cuh"
+#include "parse_utilities.cuh"
+#include "utilities/common.cuh"
+#include "utilities/error.cuh"
+#include "utilities/gpu_vector.cuh"
+#include "utilities/read_file.cuh"
+#include <iostream>
+#include <vector>
+
+__device__ void warpReduceDipole(volatile float* sdata, int tid)
```
brucefan1983, line 31: This is deprecated (and in principle wrong) way to do reduction. You can follow the style of the similar kernels I wrote (there is one in force.cu). 

Saem comment for polarizability.

src/force/potential.cu
elindgren: Ah okay, I've restored it!

src/force/force.cu
elindgren: Gotcha :+1:

src/force/nep3_small_box.cuh
elindgren: Right, didn't think of that :facepalm:

src/measure/dump_dipole.cu
elindgren: Ah okay, I see! I followed the one in force.cu, it was much shorter and neater :D

elindgren:
@brucefan1983 Thanks! I've addressed your comments.

src/force/nep3.cu
```diff
@@ -274,6 +286,10 @@ NEP3::NEP3(const char* file_potential, const int num_atoms)
 
   annmb.num_para =
     (annmb.dim + 2) * annmb.num_neurons1 * (paramb.version == 4 ? paramb.num_types : 1) + 1;
+  if (paramb.model_type == 2) {
```
brucefan1983, line 289: This missing for multi-GPU version.

src/force/nep3.cu
```diff
@@ -123,6 +123,18 @@ NEP3::NEP3(const char* file_potential, const int num_atoms)
...
+    paramb.model_type = 1;
+  } else if (tokens[0] == "nep3_polarizability") {
+    paramb.version = 3;
+    paramb.model_type = 2;
+  } else if (tokens[0] == "nep4_dipole") {
+    paramb.version = 4;
+    paramb.model_type = 1;
+  } else if (tokens[0] == "nep4_polarizability") {
+    paramb.version = 4;
+    paramb.model_type = 2;
```
brucefan1983, line 137: missing version 2 for dipole and polarizability

src/force/nep3_multigpu.cu
```diff
@@ -104,6 +104,18 @@ NEP3_MULTIGPU::NEP3_MULTIGPU(
...
+    paramb.model_type = 1;
+  } else if (tokens[0] == "nep3_polarizability") {
+    paramb.version = 3;
+    paramb.model_type = 2;
+  } else if (tokens[0] == "nep4_dipole") {
+    paramb.version = 4;
+    paramb.model_type = 1;
+  } else if (tokens[0] == "nep4_polarizability") {
+    paramb.version = 4;
+    paramb.model_type = 2;
```
brucefan1983, line 118: missing version 2 for dipole and polarizability

brucefan1983:
Looks good to me now. Thanks! Time to merge!

PR #537, opened 2023-12-23
updated examples and scripts in folder vasp2xyz 

Yanzhou-Wang:
1. Little bugs fixed in vasp2xyz shell scripts, and these scripts are more robust
2.  Rebuilt the examples in folder vasp2xyz and Shortened  the their sizes


brucefan1983:
why there are only added lines, no deleted ones?

brucefan1983:
And what bug is it? Need to describe it more celarly.

Yanzhou-Wang:
1. Both older scripts "singleFrame-outcars2nep-exyz.sh" and "multipleFrames-outcars2nep-exyz.sh"  fail to  handle correctly element symbols with subscript like Pd_d, Ti_pv, B_h, etc when building xyz-extended dataset. Now the bugs have been fixed.
2. Both updated scripts are more friendly than before.
3. Demo examples for both scripts are now re-organized more clearly.

PR #538, opened 2023-12-24
try message passing again

brucefan1983:
So far I have not seen the benefit of using 1 layer of message passing. Perhaps it requires at leat 2 layers to have some effect, but that would lead to very bad computational performance: at least one order of magnitude slower and no simple formulation of virial and heat current. These are not what I wanted. Perhpas it's time to say bye bye to message passing!


shdchen:
I feel that significant progress has been made. And there's a good chance we can find ways to simplify the framework and enhance performance. Let's continue to brainstorm and explore potential optimizations. I strongly feel that NEP5, with a combination of higher accuracy of message passing and improved efficiency, is still within reach.

PR #539, opened 2023-12-25
Initialize species count to zero also for restarts

erhart1:
This PR fixes a bug that occurs when the `mc` command is used more than once in a job script, e.g., when alternating between different ensembles.

The counter for the number of species is set up by `resize`, which initializes the elements of the vector to zero for the *first* call. When calling the `mc` command repeatedly before this PR, this caused the species counter to be incremented with every additional `mc` command, leading to incorrect species counts and concentrations. To prevent this behavior this PR inserts one line after the `resize` command that resets the counter.


brucefan1983:
good catch.

PR #541, opened 2024-01-03
better default regularization strength

brucefan1983:
The original default regularization weight parameters $\lambda_1$ and $\lambda_2$ are set equal to the combined physical RMSE, which is not always optimal. Particularly, when the initial RMSE is large, the training will simply get stuck due to too strong regularization. Based on many tests we have seen so far, it seems a good default value is $\lambda_1=\lambda_2=\sqrt{N}/1000$, where $N$ is the total number of training parameters.

### DIFF
--- a/src/main_nep/parameters.cu
+++ b/src/main_nep/parameters.cu
@@ -184,6 +184,13 @@ void Parameters::calculate_parameters()
     number_of_variables += number_of_variables_ann;
   }
 
+  if (!is_lambda_1_set) {
+    lambda_1 = sqrt(number_of_variables * 1.0e-6f);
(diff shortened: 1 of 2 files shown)


erhart1:
This choice makes a lot of sense to me in terms of the Bayesian motivation for the loss function, with the lambda parameters being the inverse variance and mean of the Gaussian (L2) and Laplace (L1) distributions that are the priors for the parameters.

Looks good to me!

shdchen:
This looks like a good choice.

shdchen (commented):
Thanks for this improvement on regularization.

PR #543, opened 2024-01-04
add comments on lambda_e

brucefan1983:
Do you agree with my comments in the doc? @erhart1


shdchen:
I've been running some tests on this loss schedule (varying lambda_f and lambda_e), and may have some more converged results in a few more days.

PR #545, opened 2024-01-14
Add piston dump.

psn417:
In this PR I did the following things:

[1] Add a TI ensemble to perform coupling constant integration at fixed lambda value. This is for testing purpose.
[2] Add a piston ensemble and dump_piston command. I've tested it and it produce good result:
image
This is a image of the pressure distribution of a two-wave structure.


brucefan1983:
Looks good to me. Thanks fo the nice work.

PR #547, opened 2024-01-16
Add Hugoniot thermostat

psn417:
I add the Hugoniot thermostat. This is similiar to MSST but you can specify the target pressure. It is useful in some cases. I've tested it on a system and it works fine.

I inherit it from MTTK so there ain't much code.


brucefan1983:
nice new features.

PR #578, opened 2024-03-27
Better default NEP hyperparameters

brucefan1983:
We should expose the *best* default hyperparameters to the users.

* Based on our recent tests, $L_1$ regularization is not needed. $L_1$ regularization can totally kill some parameters, which makes new training restarting from old one(s) difficult. It takes a long time to bring the dead parameters alive.
* Set the default values for $N_{\rm bas}^{\rm R/A}$ from (12, 12) back to (8, 8), which are actually more than enough. Using (12,12) leads to too many trainable parameters for large models with many species.

### DIFF
--- a/doc/nep/input_files/nep_in.rst
+++ b/doc/nep/input_files/nep_in.rst
@@ -83,7 +83,7 @@ Here is an example :attr:`nep.in` file using all the default parameters::
   version       4       # default
   cutoff     	8 4     # default
   n_max      	4 4     # default
-  basis_size	12 12   # default
+  basis_size	8 8     # default
   l_max      	4 2 0   # default
   neuron     	30      # default
   lambda_e      1.0     # default
--- a/doc/nep/input_parameters/basis_size.rst
+++ b/doc/nep/input_parameters/basis_size.rst
@@ -14,6 +14,6 @@ The syntax is::
 where :attr:`<N_bas_R>` and :attr:`<N_bas_A>` set :math:`N_\mathrm{bas}^\mathrm{R}` and :math:`N_\mathrm{bas}^\mathrm{A}`, respectively.
 The parameters must satisfy :math:`0 \leq N_\mathrm{bas}^\mathrm{R},N_\mathrm{bas}^\mathrm{A} \leq 19`.
 
-The default values of :math:`N_\mathrm{bas}^\mathrm{R}=12` and :math:`N_\mathrm{bas}^\mathrm{A}=12` are usually sufficient.
+The default values of :math:`N_\mathrm{bas}^\mathrm{R}=8` and :math:`N_\mathrm{bas}^\mathrm{A}=8` are usually sufficient.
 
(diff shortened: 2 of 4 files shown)


shdchen:
Yes, these hyperparameters appear to be optimal default choice according to our recent tests, with a good balance between accuracy and efficiency.

elindgren:
I actually see L1 completely removing parameters as a feature, as it gives more robust models in the sense of being less likely to overfit. 

With starting training from old ones, do you mean transfer learning? So starting the training of one model from another?

shdchen:
> I actually see L1 completely removing parameters as a feature, as it gives more robust models in the sense of being less likely to overfit.
> 
> With starting training from old ones, do you mean transfer learning? So starting the training of one model from another?
> 
> Ping @erhart1 what are your thoughts on this?

Thank you, @elindgren, for sharing your insights.

Yes, our recent tests involved multiple iterations of the NEP model training on various datasets of increasing sizes. We observed that employing L1 regularization during the initial stages might limit the model's adaptability to new datasets in subsequent stages. Therefore, in the context of multi-stage training, temporarily excluding L1 regularization could help maintain the flexibility of the early-stage 'baby' model.

However, L1 regularization could still be valuable for finalizing the NEP model training, as it could help prevent overfitting and may enhance the robustness of the ultimate 'grownup' model.

Considering that most users may not necessarily use multi-stage training, perhaps we could keep L1 regularization enabled by default?

What do you think? @brucefan1983

brucefan1983:
> I actually see L1 completely removing parameters as a feature, as it gives more robust models in the sense of being less likely to overfit.
> 
> With starting training from old ones, do you mean transfer learning? So starting the training of one model from another?
> 
> Ping @erhart1 what are your thoughts on this?

Many users want to restart a training after adding structures. We are also trying to combined two subsystems (such as one for A,B,C,D,D elemets and the other for E,F,G,H,I elemets) to train a bigger one (for A-I elements). In these cases, we found having L1 reguralization enabled in the early stage(s) can slow down the learning process. As @shdchen commented, one can add L1 regularization at the last step if that is intended. 

My major concern is: what is the best default choice that is usually adopted by a new user?

brucefan1983:
> > I actually see L1 completely removing parameters as a feature, as it gives more robust models in the sense of being less likely to overfit.
> > With starting training from old ones, do you mean transfer learning? So starting the training of one model from another?
> > Ping @erhart1 what are your thoughts on this?
> 
> Thank you, @elindgren, for sharing your insights.
> 
> Yes, our recent tests involved multiple iterations of the NEP model training on various datasets of increasing sizes. We observed that employing L1 regularization during the initial stages might limit the model's adaptability to new datasets in subsequent stages. Therefore, in the context of multi-stage training, temporarily excluding L1 regularization could help maintain the flexibility of the early-stage 'baby' model.
> 
> However, L1 regularization could still be valuable for finalizing the NEP model training, as it could help prevent overfitting and may enhance the robustness of the ultimate 'grownup' model.
> 
> Considering that most users may not necessarily use multi-stage training, perhaps we could keep L1 regularization enabled by default?
> 
> What do you think? @brucefan1983

Yes, I agree to enable L1 regularization by default. After all, the cases we tested are not for beginners.

brucefan1983:
I have kept the L1 regularization by default. @elindgren

PR #591, opened 2024-04-02
Add momentum mirror method for shock wave simulation.

psn417:
This is an alternative for `ensemble piston`.
I'm testing it.


psn417:
I've finished testing. It works fine.

PR #611, opened 2024-04-27
Update publications.bib

hailan2005:


hailan2005:
Update publications.bib

PR #614, opened 2024-05-05
allow compute SHC for all group IDs except group ID 0 in a given grouping method

Tingliangstu:
It was modified based on Dr. Xuke's version. In the original version, only the SHC of all group IDs in the grouping method 0 could be calculated. Grouping method 0 is already occupied by the heat source and sink when using NEMD. In this case, the group IDs in group method 0 cannot be divided arbitrarily.

In the PR version, there is no limit to grouping method 0. All possible group methods can be specified. **In addition, the SHC output group ID 0 is skipped**. Because when using NEMD, users tend not to output fix groups and heat sink and source groups. This means that when regrouping, the user can assign this part to group ID 0, thereby avoiding wasting computation time and memory to compute all the other remaining group IDs. Of course, when the user still wants to compute all group time for all systems, they can do so without assigning atoms to group ID 0. This provides some flexibility.

In this PR, the main modification is **shc.cu**. The corresponding user manual and output instructions have also been modified accordingly.


Tingliangstu:
@brucefan1983 @Kick-H

PR #617, opened 2024-05-06
better hyperparameters in snes

brucefan1983:
* Increasing initial means $\mu$ from the interval [0.5, 0.5] to the interval [-1, 1], which might be a more optimal starting point.
* Reducing initial searching variances $\sigma$ from 0.1 to $(3 + \ln N) / (5 \sqrt{N})$, which might make the training smoother, particularly in the beginning.
* Reducing the default regularization strength by 10 times, in accordance with the reduced $\sigma$. Now a good regularization parameter is of the order of 0.005. 

Now I understand why SNES needs so large regularization previously. It was due to the too large initial searching variance (learning rates for the parameters).

### DIFF
--- a/src/main_nep/parameters.cu
+++ b/src/main_nep/parameters.cu
@@ -193,17 +193,17 @@ void Parameters::calculate_parameters()
 
   if (version == 4) {
     if (!is_lambda_1_set) {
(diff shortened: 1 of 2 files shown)


shdchen (approved):
Looks good according to our recent tests.

hityingph (approved):

erhart1 (approved):

Issue #605, opened 2024-04-16
Generate seperate frames by dump_exyz


mushroomfire:
For large scale simulations, an entire trajectory is not easy to check and hard to download it remotely. Providing a way to save each frame can solve this problem. For example, dump.0.xyz, dump.100.xyz, dump.200.xyz ...

brucefan1983:
now fixed by your PR.

PR #629, opened 2024-05-22
Add separated parameter for dump_exyz

mushroomfire:
This PR adds a parameter in dump_exyz to generate separated frames to fix issue #605 . The documentation is also updated.


brucefan1983:
Great addition. Thanks!

PR #637, opened 2024-06-01
add PCA sampling script

hityingph:
In tools, I have added a new script to select structures from original large reference dataset, based on Calorine and farthest point sampling. I use the PbTe system as a toy example, select 25 structures from the 300 structures in text.xyz.


brucefan1983:
thanks!

PR #640, opened 2024-06-10
allow for tune normalization

brucefan1983:
Add this keyword in nep.in:

```
initial_para <value>
```

`value` can be within [0.1, 1], defaulting to 1 (as in previous master).

Reducing this value will make higher-body correlations more important and also increases the overall values of the descriptors. Therefore, one might need to increase the regularization weights (one can check the default values from screen output) after reducing `value`.


shdchen (commented):
Looks good.

PR #660, opened 2024-06-29
typewise cutoffs

brucefan1983:
* In this PR, I propose to make the NEP cutoffs  (radial and angular) and the ZBL cutoffs (outer and inner) type (species) dependent. They will be calculated based on the so-called covalent radii of the elements (taken from the DFT-D3 method):
```c++
__constant__ float COVALENT_RADIUS[94] = 
{0.426667f,0.613333f,1.6f,1.25333f,1.02667f,1.0f,0.946667f,0.84f,0.853333f,0.893333f,
  1.86667f,1.66667f,1.50667f,1.38667f,1.46667f,1.36f,1.32f,1.28f,2.34667f,2.05333f,
  1.77333f,1.62667f,1.61333f,1.46667f,1.42667f,1.38667f,1.33333f,1.32f,1.34667f,1.45333f,
  1.49333f,1.45333f,1.53333f,1.46667f,1.52f,1.56f,2.52f,2.22667f,1.96f,1.85333f,
  1.76f,1.65333f,1.53333f,1.50667f,1.50667f,1.44f,1.53333f,1.64f,1.70667f,1.68f,
  1.68f,1.64f,1.76f,1.74667f,2.78667f,2.34667f,2.16f,1.96f,2.10667f,2.09333f,
  2.08f,2.06667f,2.01333f,2.02667f,2.01333f,2.0f,1.98667f,1.98667f,1.97333f,2.04f,
  1.94667f,1.82667f,1.74667f,1.64f,1.57333f,1.54667f,1.48f,1.49333f,1.50667f,1.76f,
  1.73333f,1.73333f,1.81333f,1.74667f,1.84f,1.89333f,2.68f,2.41333f,2.22667f,2.10667f,
  2.02667f,2.04f,2.05333f,2.06667f};
```

* I also removed NEP2.

pseudo code:

```c++
rc_radial[i][j] = min(rc_radial_global,  (covalent_radius[i] + covalent_radius[j]) * 2.5)
rc_angular[i][j] = min(rc_angular_global,  (covalent_radius[i] + covalent_radius[j]) * 2.0)
rc_zbl_outer[i][j] = min(rc_zbl_global,  (covalent_radius[i] + covalent_radius[j]) * 0.6)
```

Usage example for `nep.in`:
```
cutoff 8 6 # global NEP cutoff
zbl      2    # global ZBL cutoff
use_typewise_cutoff # enable this feature for NEP cutoff
use_typewise_cutoff_zbl # enable this feature for ZBL cutoff
```
For C-C we will have
```
rc_radial = 5
rc_angular = 4
rc_zbl_outer = 1.2
```

For O-H we will have
```
rc_radial = 3.1667
rc_angular = 2.5333
rc_zbl_outer = 0.76
```

Potential benefits compared to using large global cutoffs only: 
* More accurate: there should be optimal cutoffs for different species, both for NEP and ZBL.
* More stable and data efficient: less redundant features
* More efficient: will have less neighbors


elindgren:
@brucefan1983 interesting! Can one still set a custom global cutoff? That might be nice in cases where one uses dummy species for e.g. excited states or similar in the NEP model, and want more control over the cutoffs. Maybe one can make these species-dependent cutoff the default if one does not specify a cutoff value? 

Also, would are these fixed values then? Would it make sense to have the cutoffs also trainable (like the `c_ijkl` coefficients?

brucefan1983:
> @brucefan1983 interesting! ~Can one still set a custom global cutoff? That might be nice in cases where one uses dummy species for e.g. excited states or similar in the NEP model, and want more control over the cutoffs. Maybe one can make these species-dependent cutoff the default if one does not specify a cutoff value?~ Edit: I saw that one toggles this feature on, so my comment is irrelevant 😅
> 
> Also, would are these fixed values then? Would it make sense to have the cutoffs also trainable (like the `c_ijkl` coefficients?

1）The global cutoffs are still needed, as in the expressions below:
```
rc_radial[i][j] = min(rc_radial_global,  (covalent_radius[i] + covalent_radius[j]) * 2.5)
rc_angular[i][j] = min(rc_angular_global,  (covalent_radius[i] + covalent_radius[j]) * 2.0)
rc_zbl_outer[i][j] = min(rc_zbl_global,  (covalent_radius[i] + covalent_radius[j]) * 0.7)
```

It is not easy to make these cutoffs trainable, because this will make the memory allocation too dynamic and slow down GPU calculation and complicate the implementation. Even without this problem, training them would be not a good idea I think. The idea here is to put physics/chemistry in by hand and make the training more smooth, the model more stable, and more efficient.

shdchen (approved):
Looks good. Utilizing typewise cutoffs tends to accelerate convergence and improve the accuracy of training. We tested this approach on a couple of multi-element systems, and it worked well. Thanks for this enhancement!

PR #685, opened 2024-08-06
2 link references are corrected

Yanzhou-Wang:
In the Sec. "Thermal Transport from HNEMD" of gpumd Tutorials, 2 reference links are corrected (old links are missing) for the definitions of both $K(t)$ and $\kappa(\omega)$.


Yanzhou-Wang:
Redundant cache file ".ipynb_checkpoints" was removed.

PR #690, opened 2024-08-14
add a script to shift potential energy to zero

tamaswells:


tomasfbouvier:
Hmm but it is not only about shifting the energy of your dataset. One would somehow need to force NEP to predict 0 for an isolated atom. That is not currently possible. We have a dataset where the energies have been shifted to be wrt the isolateed atom system. Yet NEP ends up learning to predict a non-zero energy for the isolated atom, which is problematic, as instead of minimising the rmsd wrt  the binding energy the optimizer is "optimising" the value of e0.

PR #691, opened 2024-08-14
add NSW check for AIMD simulations

zhyan0603:
Hi Dr.Yanzhou,

I've received a feedback regarding a minor bug in the previous script, which arose after I added the NELM check. As a result, the OUTCARs from AIMD simulations were incorrectly flagged as non-convergent.

To address this issue, I have now included an NSW check:

- If NSW is 0, it indicates a self-consistent field (SCF) calculation, following the original logic.
- If NSW is not 0, the script will directly judge the file as being from an AIMD calculation, assuming convergence.

Best,
Zihan


Yanzhou-Wang (approved):

Yanzhou-Wang:
Hi Zihan,

Thanks for your nice explanation and debugging  effort for this little bug.

Yanzhou

PR #762, opened 2024-10-22
Small Acceleration by Reduce Local Memory Operators

BBBuZHIDAO:
# Use temporary variables to reduce local memory loading and storing in `find_fn` and `find_fn_and_fnp` functions

## Analysis
When I check code in `nep_utilities.cuh` file, I find `find_fn` takes a lot of time. Detail shows below:
image
The right panel shows that the memory operators is complicated.
* the gray bar means `long scoreboard`, usually caused by global memory operators.
* the blue bar means instructions executed.

This memory chart shows memory requests and other information for `find_descriptor_angular` function.
image

## Modification
Use some temporary variables to save these local variables to make sure there is only one storing operator.
image

This memory chart is also for `find_descriptor_angular` function, setting non-modified as the baseline. Although L2 cache storing is also the peek, requests are nearly half.
image

## Because I modify two important functions, so maybe It needs be checked carefully

I have tested the code in some examples. It will take less time about **10%** to **16%**.

### DIFF
--- a/src/utilities/nep_utilities.cuh
+++ b/src/utilities/nep_utilities.cuh
@@ -360,14 +360,27 @@ static __device__ __forceinline__ void find_fn_and_fnp(
 static __device__ __forceinline__ void
 find_fn(const int n_max, const float rcinv, const float d12, const float fc12, float* fn)
 {
+  // float x = 2.0f * (d12 * rcinv - 1.0f) * (d12 * rcinv - 1.0f) - 1.0f;
+  // fn[0] = 1.0f;
+  // fn[1] = x;
+  // for (int m = 2; m <= n_max; ++m) {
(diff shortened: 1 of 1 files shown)


BBBuZHIDAO:
Dear Bruce, I tried the `-DDEBUG` flag and found there is a bug. But don't worry, the latest commit fixed it. :smiley_cat:
Here are two test log files for [2023_Ying_bilayer_graphene](https://gitlab.com/brucefan1983/nep-data/-/tree/main/2023_Ying_bilayer_graphene/PBE/4.5_4.5?ref_type=heads):
[nep.out.raw.txt](https://github.com/user-attachments/files/17479284/nep.out.raw.txt)
[nep.out.mod.txt](https://github.com/user-attachments/files/17479272/nep.out.mod.txt)
image
image

brucefan1983:
This is a great improvement!

PR #764, opened 2024-10-22
Updated documentation of "compute" command

erhart1:
This PR adds `momentum` as allowed "quantity" in the signature of the `compute` keyword.


erhart1:
@brucefan1983 the `momentum` option was not included in the documentation page for the `compute` keyword. Please confirm that this functionality is fully functional before merging.

Issue #763, opened 2024-10-22
Example 12 and Example 13 Throw nan in Total-Loss Column


BBBuZHIDAO:
Hello, I pulled the lasted code and tested the example 12 and 13 in `example` directory. The total loss column always showed `nan`. I just modified the compiling `CUDA_ARCH` to `sm_89` and used less generation to test. Are there any other steps I missed? (or maybe the examples need to update)

Output files:
[gpumd.out.example_12.txt](https://github.com/user-attachments/files/17472131/gpumd.out.example_12.txt)
[gpumd.out.example_13.txt](https://github.com/user-attachments/files/17472135/gpumd.out.example_13.txt)
![image](https://github.com/user-attachments/assets/ba60ca97-4f73-4415-9fd0-addb94c24016)

Thank you all!

brucefan1983:
Yes, this is a harmless but elusive bug that I have no idea what's going on. I have never reproduced it.

PR #770, opened 2024-10-24
Init Struct Energy for Train Mode 1 and 2

BBBuZHIDAO:
# This PR is for #763 

There is no initialization for energy if `train_mode` equals to 1 or 2.
image

When I fix it, the `Total Loss` is not `nan`.
image

## DIfferent Results (?
I use `-DDEBUG` flag to compile and compare the `loss.out` file from initialized and non initialized code. It shows that other columns are also different. The `loss.out` files are below: (I have removed nep.restart file each time)
[loss.out.raw.txt](https://github.com/user-attachments/files/17506277/loss.out.raw.txt)
[loss.out.fix.txt](https://github.com/user-attachments/files/17506279/loss.out.fix.txt)

## One More Thing
There are many ways to solve this bug. I recommend that initialize it in struct definition. And if it's allowed, make sure all variables will be initialized. This work will not make the code slower but will make it more robust.

### DIFF
--- a/src/main_nep/structure.cuh
+++ b/src/main_nep/structure.cuh
@@ -24,7 +24,7 @@ struct Structure {
   int has_virial;
   int has_temperature;
   float weight;
-  float energy;
+  float energy = 0.0f;
   float virial[6];
   float box_original[9];
   float volume;


brucefan1983:
great finding.

PR #790, opened 2024-11-19
xnet

brucefan1983:
usage in `nep.in`:

```
use_xnet # change to use the Cauchy activation function instead of tanh
```

Original activation in the paper is $f(x) = \frac{\lambda_1 x + \lambda_2}{x^2 + d^2}$, where $\lambda_1$, $\lambda_2$, and $d$ are trainable parameters.

I modified it to $f(x) = \frac{\lambda_1 x + \lambda_2}{x^2 + d^2 + 0.01}$ to avoid singularity. Is this necessary?

Ref:  [Cauchy activation function and XNet](https://doi.org/10.48550/arXiv.2409.19221)

Does not seem to be useful, will give up soon...

### DIFF
--- a/src/main_nep/nep.cu
+++ b/src/main_nep/nep.cu
@@ -248,6 +248,7 @@ NEP::NEP(
   int deviceCount)
 {
   paramb.version = version;
+  paramb.use_xnet = para.use_xnet;
   paramb.rc_radial = para.rc_radial;
   paramb.rcinv_radial = 1.0f / paramb.rc_radial;
   paramb.rc_angular = para.rc_angular;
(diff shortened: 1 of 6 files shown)


BBBuZHIDAO:
It's glad that there is a new activation function and you have achieved it so fast :star_struck:. But I have a small question for the modification of the denominator. It means `d` in non-modified equation must be bigger than `1`. This setting will loss some functions, whose **peak value** is closer to `0`.
I try to draw function $\frac{x}{x^2+d^2}$, $\lambda_1=1$, for different `d`. Here is the result:
Figure_1
Although $\lambda$ will scale the function, the inflexion point message nay loss.

brucefan1983:
So far the feedbacks are negative :(

PR #816, opened 2024-12-07
add threshold_lo_ & threshold_hi_

Youhaojen:
fix the issue "If the system has exploded, unphysical structures may be saved since no upper bound is set on the uncertainty  threshold."

**Summary**
The active learning workflow can be used with `active <interval> <has_velocity> <has_force> <threshold_lo> <threshold_hi>`

**Modification**
Rename `threshold_` to `threshold_lo_`, and add `threshold_hi`

**Others**

### DIFF
--- a/src/measure/active.cu
+++ b/src/measure/active.cu
@@ -113,8 +113,8 @@ void Active::parse(const char** param, int num_param)
   check_ = true;
   printf("Active learning.\n");
 
-  if (num_param != 5) {
-    PRINT_INPUT_ERROR("active should have 4 parameters.");
+  if (num_param != 6) {
+    PRINT_INPUT_ERROR("active should have 5 parameters.");
   }
   if (!is_valid_int(param[1], &check_interval_)) {
     PRINT_INPUT_ERROR("check interval should be an integer.");
@@ -141,13 +141,16 @@ void Active::parse(const char** param, int num_param)
   } else {
     printf("    with force data.\n");
   }
-  if (!is_valid_real(param[4], &threshold_)) {
+  if (!is_valid_real(param[4], &threshold_lo_)) {
+    PRINT_INPUT_ERROR("threshold should be a real number.\n");
+  }
+  if (!is_valid_real(param[5], &threshold_hi_)) {
     PRINT_INPUT_ERROR("threshold should be a real number.\n");
   }
 
   printf(
-    "    will check if uncertainties exceed %f every %d iterations.\n",
-    threshold_,
+    "    will check if uncertainties exceed %f and %f every %d iterations.\n",
+    threshold_lo_, threshold_hi_,
     check_interval_);
 }
 
@@ -238,7 +241,7 @@ void Active::process(
     }
   }
   write_uncertainty(step, global_time, uncertainty);
-  if (uncertainty > threshold_) {
+  if (threshold_hi_ > uncertainty && uncertainty > threshold_lo_) {
     write_exyz(
       step,
       global_time,


brucefan1983:
Thanks for the contribution. It seems you separated changes into two pull requests? Could you complete one and delete the other?

Youhaojen:
Dear Prof. Fan,

Thanks for your reply. One is for modifying active.cu another one is for modifying activate.cuh. I’m not sure how to delete another pull request. Could you merge the two in one pull request?

Thanks

Hao-Jen You

brucefan1983:
OK, I will take care of it. I need to study the code more carefully when I have enough time. I will contact you if I have more questions.

src/measure/active.cu
```diff
@@ -238,7 +241,7 @@ void Active::process(
     }
   }
   write_uncertainty(step, global_time, uncertainty);
-  if (uncertainty > threshold_) {
+  if (threshold_hi_ > uncertainty && uncertainty > threshold_lo_) {
```
brucefan1983, line 244: This does not seem be correct to me.

brucefan1983:
@elindgren Do you think it is necessary to have an upper limit?

src/measure/active.cu
```diff
@@ -141,13 +141,16 @@ void Active::parse(const char** param, int num_param)
...
+  }
+  if (!is_valid_real(param[5], &threshold_hi_)) {
     PRINT_INPUT_ERROR("threshold should be a real number.\n");
   }
 
   printf(
-    "    will check if uncertainties exceed %f every %d iterations.\n",
-    threshold_,
+    "    will check if uncertainties exceed %f and %f every %d iterations.\n",
+    threshold_lo_, threshold_hi_,
```
elindgren, line 153:
```suggestion
    "    will check if uncertainties exceed %f and are below %f every %d iterations.\n",
    threshold_lo_, threshold_hi_,
```

src/measure/active.cu
```diff
@@ -141,13 +141,16 @@ void Active::parse(const char** param, int num_param)
   } else {
     printf("    with force data.\n");
   }
-  if (!is_valid_real(param[4], &threshold_)) {
+  if (!is_valid_real(param[4], &threshold_lo_)) {
+    PRINT_INPUT_ERROR("threshold should be a real number.\n");
+  }
+  if (!is_valid_real(param[5], &threshold_hi_)) {
```
elindgren, line 147:
```suggestion
  if (!is_valid_real(param[4], &threshold_low_)) {
    PRINT_INPUT_ERROR("threshold_low should be a real number.\n");
  }
  if (!is_valid_real(param[5], &threshold_high_)) {
```

src/measure/active.cu
```diff
@@ -141,13 +141,16 @@ void Active::parse(const char** param, int num_param)
...
     PRINT_INPUT_ERROR("threshold should be a real number.\n");
```
elindgren, line 148:
```suggestion
    PRINT_INPUT_ERROR("threshold_high should be a real number.\n");
```

src/measure/active.cu
elindgren, line 244:
```suggestion
  if (threshold_high_ > uncertainty && uncertainty > threshold_low_) {
```

elindgren (changes requested):
I think it makes sense to have an upper cutoff for the use case that you describe. The high cutoff could potentially be an internal value, but it would probably be difficult to know what an appropriate value would be in general.

For readability, I would suggest renaming `threshold_lo_` to `threshold_high_` and `threshold_hi_` to `threshold_high_`, respectively.

Also, modify the documentation for the `active` keyword to reflect the new usage, including the examples. The relevant file can be found here: https://github.com/brucefan1983/GPUMD/blob/master/doc/gpumd/input_parameters/active.rst

elindgren:
> @elindgren Do you think it is necessary to have an upper limit?

I have not encountered this situation personally, but I can see it potentially being an issue. My main concern is if it should be set by the user, or be an internal fixed parameter. As I said in my other comment, I think it's difficult to set a good value for an internal parameter however.

Youhaojen:
@elindgren I agree. The value of the upper cutoff sometimes requires some experience to choose. However, this functionality mainly references the [exploration part of dpgen](https://docs.deepmodeling.com/projects/dpgen/en/latest/run/example-of-param.html#exploration). Therefore, I think we might be able to implement this functionality here as well.

BTW, don't forget to modify the `active.cuh` together. #817

brucefan1983:
> @elindgren I agree. The value of the upper cutoff sometimes requires some experience to choose. However, this functionality mainly references the [exploration part of dpgen](https://docs.deepmodeling.com/projects/dpgen/en/latest/run/example-of-param.html#exploration). Therefore, I think we might be able to implement this functionality here as well.
> 
> BTW, don't forget to modify the `active.cuh` together. #817

I think you need to take care of this. We cannot merge a broken code.

brucefan1983:
Do you still want to finishe this PR? @Youhaojen

Youhaojen:
The PR can be closed. This is not necessary for GPUMD, as the function has already been added to my active learning workflow code. The code is almost completed, and it will also support GPUMD 4.0.0. I will let you know once the code is released. Thanks. @brucefan1983

brucefan1983:
> The PR can be closed. This is not necessary for GPUMD, as the function has already been added to my active learning workflow code. The code is almost completed, and it will also support GPUMD 4.0.0. I will let you know once the code is released. Thanks. @brucefan1983

OK, I am looking forward to your new work!

PR #817, opened 2024-12-07
add threshold_lo_ & threshold_hi_

Youhaojen:
fix the issue "If the system has exploded, unphysical structures may be saved since no upper bound is set on the uncertainty  threshold."

**Summary**
The active learning workflow can be used with `active <interval> <has_velocity> <has_force> <threshold_lo> <threshold_hi>`

**Modification**
Rename `threshold_` to `threshold_lo_`, and add `threshold_hi`

**Others**

### DIFF
--- a/src/measure/active.cuh
+++ b/src/measure/active.cuh
@@ -44,7 +44,8 @@ private:
   int check_interval_ = 1;
   int has_velocity_ = 0;
   int has_force_ = 0;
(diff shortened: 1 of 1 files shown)


brucefan1983:
Thanks for the contribution. It seems you separated changes into two pull requests? Could you complete one and delete the other?

PR #832, opened 2024-12-24
output descriptor

brucefan1983:
# New feature

Add option to output descriptor during prediction.

# Usage

```
# Both should be 1 to enable this feature:
prediction        1 
output_descriptor 1 
```

Results will be output to `descriptor.out` (with appending mode). The number of rows is the number of atoms in `train.xyz`, with the same order. The number of columns is the dimension of the descriptor.

### DIFF
--- a/doc/nep/input_parameters/index.rst
+++ b/doc/nep/input_parameters/index.rst
@@ -34,3 +34,4 @@ Below you can find a listing of keywords for the ``nep.in`` input file.
    batch
    population
    generation
+   output_descriptor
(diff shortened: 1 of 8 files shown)


brucefan1983:
@elindgren users have confirmed that the results are consistent with calorine.

shdchen:
I just tested it, and it passed the per-atom descriptor test for the PbTe example (test.xyz). Please see the plots for comparison. Calorine (CPU) and NEP_prediction (GPU) give consistent results. 
Descriptor_comparison_Calorine_vs_NEP_prediction

shdchen (approved):
Looks good.

PR #834, opened 2024-12-26
fix: correct nframes calculation in dp2xyz.py when multiple set.* directories exist

shiruosong:
Fixed an issue where the nframes count only reflected the value from the last "set.*" directory in cases with multiple such directories.

### DIFF
--- a/tools/dp2xyz/dp2xyz.py
+++ b/tools/dp2xyz/dp2xyz.py
@@ -76,9 +76,11 @@ def to_system_data(folder):
     all_eners = []
     all_forces = []
     all_virs = []
+    all_nframes = 0
     for ii in sets:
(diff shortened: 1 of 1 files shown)


Kick-H (commented):
There is no problem with the modification here.
Previously, the data['frames'] value was hardly used, so this problem could not be discovered

shiruosong:
> There is no problem with the modification here. Previously, the data['frames'] value was hardly used, so this problem could not be discovered

Thank you for the feedback!  The issue became apparent because during the conversion to XYZ format files, only data['frames'] frames were written, while the actual data contained more frames.  This discrepancy was caused by data['frames'] only considering the last "set. *" directory instead of summing across all of them.

brucefan1983:
thanks for the improvement!

PR #840, opened 2024-12-30
Add box optimization during energy minimization for the FIRE method.

mushroomfire:
**Summary**
This PR add box optimization during minimization for thr FIRE method.

**Usage**

minimize fire 1.0e-5 1000 1 

- with hydrostatic strain

minimize fire 1.0e-5 1000 1 1

### DIFF
--- a/src/main_nep/fitness.cu
+++ b/src/main_nep/fitness.cu
@@ -19,6 +19,7 @@ Get the fitness
 
 #include "fitness.cuh"
 #include "nep.cuh"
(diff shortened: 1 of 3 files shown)


brucefan1983:
Thanks for implementing this very desired feature. I suggest you make this feature as an independent class such as class 

```C++
class Minimizer_FIRE_Varying_Box : public Minimizer
```

It is better to keep the the different versions as independent as possible, even though there are repeating lines of code.

PR #844, opened 2025-01-02
no SM less than 60

brucefan1983:
**Summary**

Starting from GPUMD-4.0, we will not support GPUs with compute capability < 6.0.

**Modification**

**Others**

### DIFF
--- a/src/force/nep_small_box.cuh
+++ b/src/force/nep_small_box.cuh
@@ -19,21 +19,6 @@
 #include "utilities/nep_utilities.cuh"
 #include "utilities/gpu_macro.cuh"
 
-#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ < 600)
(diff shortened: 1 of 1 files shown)


Dankomaister:
@brucefan1983 the Nvidia K80 GPU has compute capability 3.7 would be great if we could still support it.
The K80 is often found in legacy HPC systems and we have around 80x K80 GPUs in our local cluster which we just a lot for NEP training and running GPUMD.

brucefan1983:
> @brucefan1983 the Nvidia K80 GPU has compute capability 3.7 would be great if we could still support it. The K80 is often found in legacy HPC systems and we have around 80x K80 GPUs in our local cluster which we just a lot for NEP training and running GPUMD.

OK, I didn't realized this. I will bring it back, but perhaps requires an extra macro to be set in makefile. The reason for deleteding this is that some other cards (called DCU is not compatible with this part)

PR #845, opened 2025-01-03
Add the header file "#include <cstring>"

hailan2005:
**Summary**
The addition of the header file `#include <cstring>` is capable of resolving the error messages that pop up during compilation with NVCC in DCU.

**Modification**
The header file `#include <cstring>` has been added to all *.cu files.

**Others**


hailan2005:
The warnings that appeared when compiling with NVCC in DCU have been modified.

Issue #757, opened 2024-10-15
Unexpected error with the comment symbol # in "run.in" file.


Yanzhou-Wang:
I encountered with unexpected error with the comment symbol # in "run.in"

Lately I used GPUMD code to run relaxation of  lattice constant of graphene at finite temperature. When the "run.in" file reads as follows:
```
potential  C.nep
time_step   1
ensemble  pimd  128 75 75 100 0 0 0 1000 1000 1 1000  #ensemble pimd <num_beads> <T_1> <T_2> <T_coup> {<pressure_control_parameters>} 
dump_thermo 1000
run         200000
```
It would prompt the following error in log file:
```
Input Error:
 File:       integrate/integrate.cu
 Line:       666
  Error text: Cannot use triclinic box with only 3 target pressure     components.
```
Interestingly, the error disappears and the optimization job recovers to work normally when I remove the comments "#ensemble pimd <num_beads> <T_1> <T_2> <T_coup> {<pressure_control_parameters>} " in "run.in" file.

brucefan1983:
fixed in #854

Issue #755, opened 2024-10-15
Simulation unstable when using triclinic updates in MTTK integrators


rashidrafeek:
I was testing out some simulations with GPUMD and I noticed that the when using the `tri` option in the [`npt_mttk` ensemble](https://gpumd.org/dev/gpumd/input_parameters/ensemble_mttk.html#kw-ensemble-mttk), the atoms breaks apart. This only occurs for the `tri` option and not the `iso` or `aniso` options to update cell vectors. I initially noticed this for a model I had trained, but I also tested this with one of the models in nep-data gilab repo. I specifically used the [model for CsPbBr3](https://gitlab.com/brucefan1983/nep-data/-/blob/main/2023_Shi_CsPbX(X=Cl,Br,I)/MD_simulation_inputs/Br/nep.txt?ref_type=heads) there and the same model.xyz there, with the following input:

```
potential      nep.txt

# The equilibration stage
ensemble npt_mttk temp 0.10 0.10 tri 0 0
time_step    1
#每10000步输出一个数据
dump_thermo 100
dump_position 100
run 10000
```
The cell parameter changes indicating large volume change in the simulation is shown below: 
![cell_plot_tri](https://github.com/user-attachments/assets/39e48ec0-aff3-4deb-a2d9-462d408bc3e5)
For comparison, the following is obtained when I use `aniso` instead of `tri`:
![cell_plot_aniso](https://github.com/user-attachments/assets/eb6e4a0c-21c3-46b6-9a0c-4fb38208fec7)

brucefan1983:
I guess it is becaue the MTTK barostat is not very stable. 

So you want to control all the 6 degrees of freedom. In this case you can try to use the more stable Berendsen or SCR method. For example, using the SCR method, you can write 

```
ensemble npt_scr 0.10 0.10 100 0 0 0 0 0 0 100 100 100 100 100 100 1000
```

Also, it is a good practice to write `velocity` in the beginning to set up the initial temperature. In your example where the target temperatue is 0.1, it is better to add 

```
velocity 0.1
```

With the above considerations, you will have:

```
potential      nep.txt
velocity 0.1

# The equilibration stage
ensemble npt_scr 0.10 0.10 100 0 0 0 0 0 0 100 100 100 100 100 100 1000
time_step    1
#每10000步输出一个数据
dump_thermo 100
dump_position 100
run 10000
```

rashidrafeek:
[...]

PR #854, opened 2025-01-06
Simplify box code

brucefan1983:
**Summary**

* Remove the `int triclinic` variable in the `Box` class and simplify coding. 

* The downside is that fast potentials such as Tersoff will lose about 25% speed when the box is orthogonal.

* But for triclinic box, there is even a little gain of speed for any potential due the absence of an `if`.

**Modification**

* The doc for `thermo.out` has been updated.

**Others**

These bugs have been solved: #757 #755

**Future work**

* Examples/tutorials will be updated in a separate PR.


psn417 (approved):
I've tested MTTK and MSST. They work.

PR #858, opened 2025-01-08
Local relaxation

dongxianchao:
add a new command
mc local_minimize <mc_trials> <T> <scale_factor> <force_tolerance> <maximal_number_of_relaxation_steps> [group <grouping_method> <group_id>]
This can perform simple mc and local relaxation.

Previously, we need to write a python jobscript to do the work, randomly select two atoms and swap - relax the structure - get the energy difference to calculate the probability of swap. 
Now the work flow has been implemented in GPUMD, although there still need some modification. The local relaxation method has been implemented, which means each time swap 2 atoms, we don't need to relax the whole system to get the energy, and the volumn of relax area can be controlled by the scale factor. The sphere area with radius scale_factor * radial_cutoff will be calculated.

TO DO: modify the output file
modify the code structure

### DIFF
--- a/src/main_gpumd/run.cu
+++ b/src/main_gpumd/run.cu
@@ -36,7 +36,6 @@ Run simulation according to the inputs in the run.in file.
 #include "utilities/gpu_macro.cuh"
 #include "utilities/read_file.cuh"
 #include "velocity.cuh"
-#include <cstring>
 
 static __global__ void gpu_find_largest_v2(
   int N, int number_of_rounds, double* g_vx, double* g_vy, double* g_vz, double* g_v2_max)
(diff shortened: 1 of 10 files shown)


brucefan1983:
Good feature. Could you try to do `git merge master`? The `triclinic` member in the `Box` class has been removed. You need to modify your branch according to this change in master.

brucefan1983:
The code does not compile.
image

brucefan1983:
And the current code is not consistent with master. You need to first update master by e.g., `git pull` in the master branch.

brucefan1983:
You can check the "Files changed" carefully and confirm if all the changes are intended.

dongxianchao:
I merged the master but some files just not change somehow. The triclinic still exists.

PR #861, opened 2025-01-08
add a new command

dongxianchao:
mc local_minimize mc_trials scale_factor force_tolerance maximal_number_of_relaxation_steps [group grouping_method group_id]
This can perform simple mc and local relaxation.

Previously, we need to write a python jobscript to do the work, randomly select two atoms and swap - relax the structure - get the energy difference to calculate the probability of swap.
Now the work flow has been implemented in GPUMD, although there still need some modification. The local relaxation method has been implemented, which means each time swap 2 atoms, we don't need to relax the whole system to get the energy, and the volumn of relax area can be controlled by the scale factor. The sphere area with radius scale_factor  radial_cutoff will be calculated.
In mcmd.out file, the Maximum displacement, Average displacement and Accept ratio will be calculated. The two displacement value are calculated from the atoms falls in the range from (scale_factor - 1) * cutoff_radius to scale_factor * cutoff_radius, where is the edge of the sphere we are going to relax. One can adjust the scale factor to a reasonable value according to this. The smaller they are, the smaller the energy difference with the global relaxation will be.

TO DO: modify the output file
modify the code structure

### DIFF
--- a/src/minimize/minimizer_fire.cuh
+++ b/src/minimize/minimizer_fire.cuh
@@ -50,4 +50,15 @@ public:
     GPU_Vector<double>& potential_per_atom,
     GPU_Vector<double>& force_per_atom,
     GPU_Vector<double>& virial_per_atom);
+
+  void compute_local(
+    Force& force,
+    Box& box,
+    GPU_Vector<double>& position_per_atom,
+    GPU_Vector<int>& type,
+    std::vector<Group>& group,
+    GPU_Vector<double>& local_flags,
+    GPU_Vector<double>& potential_per_atom,
+    GPU_Vector<double>& force_per_atom,
+    GPU_Vector<double>& virial_per_atom);
 };
--- a/src/main_gpumd/run.cu
+++ b/src/main_gpumd/run.cu
@@ -480,7 +480,14 @@ void Run::parse_one_keyword(std::vector<std::string>& tokens)
   } else if (strcmp(param[0], "add_efield") == 0) {
     add_efield.parse(param, num_param, group);
   } else if (strcmp(param[0], "mc") == 0) {
-    mc.parse_mc(param, num_param, group, atom);
+    if (strcmp(param[1], "local_minimize") == 0)
+    {
+      mc.parse_mc_local(param, num_param, group, atom, box, force);
+    }
+    else
+    {
+      mc.parse_mc(param, num_param, group, atom);
+    }
   } else if (strcmp(param[0], "dftd3") == 0) {
     // nothing here; will be handled elsewhere
   } else if (strcmp(param[0], "compute_lsqt") == 0) {
--- a/src/mc/mc.cu
+++ b/src/mc/mc.cu
@@ -331,3 +331,36 @@ void MC::parse_mc(const char** param, int num_param, std::vector<Group>& groups,
 
   do_mcmd = true;
 }
+
+
+void MC::parse_mc_local(const char** param, int num_param, std::vector<Group>& group, Atom& atom, Box& box, Force& force)
+{
+  printf("Perform simple MC with local relaxation:\n");
+  int num_param_before_group = 7;
+  if (num_param > num_param_before_group)
+  {
+    parse_group(param, num_param, group, num_param_before_group);
+    printf("    only for atoms in group %d of grouping method %d.\n", group_id, grouping_method);
+  }
+  is_valid_int(param[2], &num_steps_mc);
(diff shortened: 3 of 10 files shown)


brucefan1983:
This is a very specialized feature, so I suggest you make your code localized. You should nearly not touch the existing file, but just create new files.

dongxianchao:
> This is a very specialized feature, so I suggest you make your code localized. You should nearly not touch the existing files, but just create new files.

How about write a new key word like mc_local, and construct that class totally independent of mc, but still depend on nep_energy? The minimize part, I write a new function in minimize_fire which can only update the labelled atoms by input an array like [0, 0, 1, 1, 0......], 1 means the atoms need to update and 0 means not change. I'm not sure if it's useful for other function, if it is we may keep it in minimize to increase code reusability.

brucefan1983:
You can create a new folder `mc_minimize/` and put all the new code there. The function added to the fire minimizer class is ok.

brucefan1983:
`mc_minimize` is at the same level of `mc`

brucefan1983:
Good change. Now you need to resotre the residual changes in the files within the `mc/` folder. 

You can view the changes online. For example:

image

dongxianchao:
Hi Prof. Fan,

I've finished the major part work. I add a new key word "mc_minimize", and you can use command
mc_minimize global mc_trials force_tolerance maximal_number_of_relaxation_steps [group grouping_method group_id]
for a simple MC with global relaxation. For local relaxation, here is the command
mc_minimize local mc_trials force_tolerance maximal_number_of_relaxation_steps scale_factor [group grouping_method group_id].

Now this part is highly modular. It's independent of nep_energy in mc, instead, I use the potential.cuh directly to calculate the force and energy. It's not that beautiful in minimize part. The code still use Minimizer_fire object directly. Later I may try to modify this part if there is any requirement. I feel it's better to use the interface of minimizer. But till now, the fire method seems enough for relaxation.

brucefan1983:
> Hi Prof. Fan,
> 
> I've finished the major part work. I add a new key word "mc_minimize", and you can use command mc_minimize global mc_trials force_tolerance maximal_number_of_relaxation_steps [group grouping_method group_id] for a simple MC with global relaxation. For local relaxation, here is the command mc_minimize local mc_trials force_tolerance maximal_number_of_relaxation_steps scale_factor [group grouping_method group_id].
> 
> Now this part is highly modular. It's independent of nep_energy in mc, instead, I use the potential.cuh directly to calculate the force and energy. It's not that beautiful in minimize part. The code still use Minimizer_fire object directly. Later I may try to modify this part if there is any requirement. I feel it's better to use the interface of minimizer. But till now, the fire method seems enough for relaxation.

How about speed, fast enough for you?

src/mc_minimize/mc_minimizer_local.cu
```diff
@@ -0,0 +1,766 @@
...
+  get_displacement<<<(local_N - 1) / 64 + 1, 64>>>(
+    box,
+    global_position,
+    local_position,
+    global_N,
+    local_N,
+    local_index,
+    outer_atoms_flags,
+    displacement.data());
+  gpuDeviceSynchronize();
```
brucefan1983, line 457: gpuDeviceSynchronize() can slowdown code. Use this only if it is needed. Also wrap it within CHECK()

brucefan1983:
A final question: do you want to document this feature now or leave if for the future, such as when you have a paper submitted to arxiv.

dongxianchao:
I'm not sure, let me ask Shunda.

src/minimize/minimizer_fire.cuh
```diff
@@ -50,4 +50,20 @@ public:
     GPU_Vector<double>& potential_per_atom,
     GPU_Vector<double>& force_per_atom,
     GPU_Vector<double>& virial_per_atom);
+
+
+/*
+the additional array, local_flags is needed. The length of the array should be same with atom total number.
```
brucefan1983, line 56: comment that this extra function is use for the `mc_minimize` module

dongxianchao:
> I've updated makefile.hip. The main calculation cost lies in minimize algorithm. I tested a 640000 atoms system with box length 117A, and if we use a radius of 21 A for local relaxation sphere, the speed can be 5 times faster. And the major benefit is that, the relaxation is localized, so with the increase of the system, there will be little increase in running time for same trial steps. With this feature we can calculate very large system efficiently.

I typed it wrong it's 64000 atoms

brucefan1983:
perhaps we can close this PR now?

dongxianchao:
Sure

PR #876, opened 2025-01-18
output 9 virial components via compute

### LINKED ISSUE #779: Request for Enhancement: Degree of Freedom Control and Atomic Shear Stress Output
I am currently conducting research related to dislocations and need to apply pure shear forces at the boundaries while studying the local shear stress within the simulation domain. I would like to request the developers to consider adding the following two features：
1. **An enhancement to the _fix_ parameters to include a degree of freedom selection feature**, which would enable users to independently control the degrees of freedom for a specific group of atoms in designated directions (such as x, y, or z).
2. **An extension of the _compute_ parameters to provide the capability to output atomic shear stress.** Currently, only the diagonal parts of the virial matrix, which represent normal stress, are available. The inclusion of shear stress outputs would provide a more comprehensive understanding of material behavior, particularly in the context of dislocations and defects.

brucefan1983:
**Summary**
Output 9 virial components via the `compute` keyword in `run.in`.

**Modification**

Previously it only outputs 3 diagonal components. Now the full $3\times 3$ tensor is output. So the `comput.out` file is changed.

**Others**

Partially resolve #779

### DIFF
--- a/doc/gpumd/output_files/compute_out.rst
+++ b/doc/gpumd/output_files/compute_out.rst
@@ -19,7 +19,7 @@ Assuming that the system is divided into :math:`M` groups according to the group
       
       fx_1 ... fx_M fy_1 ... fy_M fz_1 ... fz_M
 
-  * if virial is computed, there are :math:`3M` columns of group virials (in units of eV) from left to right in a form similar to that for force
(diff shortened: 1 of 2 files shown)


elindgren:
@brucefan1983 Sorry for missing to review this. I think it looks good and is a good change! :+1:

PR #890, opened 2025-01-27
Add the calculation of the minimum interatomic distance

tang070205:
Add a new file `check_distance.cu` to the model folder to check if there are any two atoms with a spacing less than 1 angstrom in the `model.xyz` file at the beginning of the calculation.

For orthogonal lattices, firstly, using the example of hnemd, the original `model.xyz` file will not report errors, but it will print out which two atoms are closest to each other normally. The atomic number starts from index 0 and serves only as a prompt, as shown in the following figure.
image

Then, I changed the coordinates of atom 0 to `0 0 0`, so that it differs from atom 1 only by 0.71 angstroms in the y direction. This will result in an error message as shown in the following figure and prompt as an input error.
image

Finally, there are periodic boundary conditions, as shown in the figure below. The left figure represents a non periodic boundary condition in the z-direction, while the right figure represents a periodic boundary condition. Non periodic boundary conditions do not generate errors, while periodic boundary conditions result in `3.35=0` in the z-direction, which is the same as the situation in the above figure.
image

For non orthogonal lattices, I used a two-dimensional material that I had manually rubbed myself, and applied a vacuum layer of 20 angstroms to the original cell.
Similarly, a normal structure will not produce errors, only prompts. You will find `atom 1 Si` and `atom 4 N`, where `1` and `4` start from the third line index of 0 in the `model.xyz` file, line by line, from `0` to `N-1` (N is the total number of atoms in the structure)
image

Next, we will still use these two atoms. I want to change the xy coordinate of N atom to be the same as that of Si atom, and there will be an error message after execution.
image

Finally, it is still a periodic issue, due to the presence of a large vacuum layer in the z-direction, I applied non periodic boundary conditions in the x-direction and doubled the expansion in the x-direction.Here, two Mo atoms are used as a demonstration, and the abscissa of the last Mo atom is added by 2 angstroms.
image

I also tested a structure with 10 million atoms, which only takes about ten seconds. From the initial screen printing to the red box in the picture, it takes about half a minute. However, when checking the atomic distance, it only takes a little over ten seconds.
image

### DIFF
--- a/src/model/check_distance.cu
+++ b/src/model/check_distance.cu
@@ -17,127 +17,150 @@
 Calculate the distance between any two atoms in the model.xyz file.
 ------------------------------------------------------------------------------*/
 
-#include "check_distance.cuh"
-#include "utilities/error.cuh"
 #include "atom.cuh"
 #include "box.cuh"
-#include <cmath>
+#include "check_distance.cuh"
+#include "utilities/error.cuh"
 #include <array>
+#include <cmath>
 #include <vector>
 
-void calculate_min_atomic_distance(const Atom& atom, const Box& box) {
-    const int N = atom.number_of_atoms;
-    const double* pos = atom.cpu_position_per_atom.data();
+void calculate_min_atomic_distance(const Atom& atom, const Box& box)
+{
+  const int N = atom.number_of_atoms;
+  const double* pos = atom.cpu_position_per_atom.data();
 
-    double min_distance = 5.0;
-    int min_i = -1, min_j = -1;
+  double min_distance = 5.0;
+  int min_i = -1, min_j = -1;
 
-    double Lx = sqrt(box.cpu_h[0]*box.cpu_h[0] + box.cpu_h[3]*box.cpu_h[3] + box.cpu_h[6]*box.cpu_h[6]);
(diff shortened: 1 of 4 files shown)


brucefan1983:
This is too slow for big systems.

brucefan1983:
You need to study the MD book more carefully :)

tang070205:
Ok, I'll learn from the MD book and change it to cuda.

brucefan1983:
> Ok, I'll learn from the MD book and change it to cuda.

It is not about CUDA, but about $O(N)$ vs $O(N^2)$ algorithms. For a system with $10^8$ atoms, even CUDA cannot make it fast enough with an $O(N^2)$ algorithm.

src/model/check_distance.cu
```diff
@@ -0,0 +1,181 @@
...
+
+void calculate_min_atomic_distance(const Atom& atom, const Box& box)
+{
+  const int N = atom.number_of_atoms;
+  const double* pos = atom.cpu_position_per_atom.data();
+
+  double min_distance = 5.0;
+  int min_n1 = -1, min_n2 = -1;
+
+  double thickness[3];
```
brucefan1983, line 79: The thicknesses are not correctly corrected.

PR #916, opened 2025-02-16
Refactor measure

brucefan1983:
**Summary**

* Refactor the measurement part to make the code more readable.
* Also prepare for enabling multiple instances of a `compute`/`dump` within one `run`.
* This PR is pure refactoring, which **should not** change any functionality, except that using multiple instances of a `compute/dump` within a run is forbidden and will trigger error message reporting.
* Compiling errors due to the missing `triclinic` member in `Box` class for NETCDF and PLUMED packages should have been fixed.

**Future work**
* Will try to enable multiple instances for some `sompute/dump` within a run.


brucefan1983:
@elindgren This is a big refactoring. Could you check the tests you created still pass?

elindgren:
> @elindgren This is a big refactoring. Could you check if the tests you created still pass?
> 
>     * dump_dipole
> 
>     * dump_polarizability
> 
>     * active
> 
>     * dump_observer
> 
> 
> @psn417 Could you check if `dump_shock_nemd` still work as intended?

@brucefan1983 Yes, I'll run them and check 👍

elindgren:
@brucefan1983 It seems the following tests fail on master. I'll run a `git bisect` later to figure out when they failed. 

image

brucefan1983:
> @brucefan1983 It seems the following tests fail on master. I'll run a `git bisect` later to figure out when they failed. Currently checking this branch as well to see if any other tests fail.
> 
> image
> EDIT: Different tests fail on this branch compared to master, so I suspect this refactor has changed things. Will give a more up-to-date comparison once I've figured out where the tests on master changed.
> 
> image

Thanks for the tests. Are there any bugs detected?

elindgren:
> > @brucefan1983 It seems the following tests fail on master. I'll run a `git bisect` later to figure out when they failed. Currently checking this branch as well to see if any other tests fail.
> > image
> > EDIT: Different tests fail on this branch compared to master, so I suspect this refactor has changed things. Will give a more up-to-date comparison once I've figured out where the tests on master changed.
> > image
> 
> Thanks for the tests. Are there any bugs detected?

I think there might be, but I need to rerun this branch after the integration tests have been fixed on master. I've made a PR with the fixed integration tests, see #933. There I also try to add the integration tests to the CI. however, we can't add the tests yet since we need a GPU runner to run them.

elindgren:
@brucefan1983 Only one of the integration tests fail after the refactoring, `test_active_dump_observer_average`. 

image

the test checks that dump_observer should not affect the active learning cycle. However, I do not think the test makes sense, since the first run is run using the PES for the first NEP in the input list, whilst the second one on the average PES. It makes sense that the structures encountered should be slightly different, and thus have slightly different uncertainties.

I think we can probably disregard the failing test. I have changed it to instead check with and without dump_observer.

brucefan1983:
Thanks for the thorough tests. I will read through the code again and
merge. Then I can start enabling multiple instances of some dump or compute
within a run. That is the major purpose of this refactoring.

Eric Lindgren ***@***.***>于周五 28.2.202516:39写道：

PR #932, opened 2025-02-27
add fine_tune command

brucefan1983:
This is a work in progress. The foundation model will be published later; it is under internal evaluation.

```
# set your new types (can be any subset of 89 elements in any order):
type       2 Te Pb


# These MUST be the same as in the foundation model:
version    4
zbl        2
cutoff     6 5
use_typewise_cutoff 4 3
use_typewise_cutoff_zbl 0.65
n_max      8 8
basis_size 8 8
l_max      6 2 1
neuron     50

# These can be different from the foundation model:
lambda_1   0
batch      1000
population 50
generation 1000 # very cheap!
# usually 1000-10000 steps are sufficient for fine-tuning; too many steps might cause forgetting existing knowledge
# It is a good idea to have a test set with diverse structures to monitor the fine-tuning process

# provide the restart file for fine tuning:
fine_tune  <path/to/foundation_model_nep.txt>  <path/to/foundation_model_nep.restart>
```


src/main_nep/fitness.cu
```diff
@@ -162,11 +162,7 @@ void Fitness::compute(
         para,
         dummy_solution.data(),
         train_set[n],
-#ifdef USE_FIXED_SCALER
-        false,
-#else
-        true,
-#endif
```
elindgren, line 169: Why drop the fixed scaler? What was it used for before?

src/main_nep/snes.cu
```diff
@@ -112,6 +117,82 @@ void SNES::initialize_mu_and_sigma(Parameters& para)
...
 }
 
+void SNES::initialize_mu_and_sigma_finetune(Parameters& para)
+{
+  // read in the whole foundation file first
+  const int num_ann_per_element = (81 + 2) * 50;
+  const int num_ann = 89 * num_ann_per_element + 1;
+  const int num_cnk_radial = 89 * 89 * 81;
+  const int num_cnk_angular = 89 * 89 * 81;
+  const int num_tot = num_ann + num_cnk_radial + num_cnk_angular;
```
elindgren, line 127: This means that the architecture of the foundation model is hard coded, right? Would it make sense to make this dynamically parse the nep.txt for the foundation model in order to potentially change it down the line?

elindgren (commented):
Very exciting! I left some comments

src/main_nep/fitness.cu
brucefan1983: To perfectly restart a training, the descriptor normolizaiton should not be changed; otherwise the descriptors might change dramatically if the dataset has changed a lot, leadding to abrupt change of optimization direction. The macro `USE_FIXED_SCLAER` was used to enforce unchanged descriptor normalization, but it was used by myself only beefore (for training the foundation model). Now it is used for restarting from the foundation model.

elindgren: Okay, makes sense!

src/main_nep/snes.cu
brucefan1983: yes, the current foundation model has a fixed arthitecture. I will modify the code to read in the nep.txt of the foundation model and check parameter consistency, report errors if some architecture parameters in `nep.in` are not the same as the foundation model.

PR #860, opened 2025-01-08
A hybrid potential "NEP+ILP" for TMDs

BBBuZHIDAO:
**Summary**
This PR is to commit a hybrid potential, NEP+ILP. The potential could simulate 2D materials: TMDs ($\mathrm{MoS_2}$, $\mathrm{MoSe_2}$, $\mathrm{WS_2}$, $\mathrm{WSe_2}$ ...) and their heterostructures. Users just need to set a group method to identify the molecular layers and a sublayer group method to identify the sublayer of each molecule, and then the interlayer potential will be set as ILP and intralayer potential will be set as NEP.

**Modification**
- Add two potential files, **ilp_nep_tmd.cu** and **ilp_nep_tmd.cuh**.
- Check this potential in **force.cu**.

**Usage**
- Use two group methods to identify the layers and sublayers, and add these group methods to the second line of ILP potential. For example, if you use group method `0` to identify the molecular layers and method `1` for sublayer. Assume you just simulate $\mathrm{MoS_2}$, then the ILP potential file should be:
```
ilp_nep_tmd 2 Mo S
0 1
5.57945037  9.37766246  2.02722246  144.15177543  97.97857017 89.43759676 2.05903055  5.12205490  491850.31619468 1.0 4.0   16.0
3.62715212  19.97137510 7.58503123  76.10193100 3.31749559  45.72032846 0.94747037  4.41042503  150597.85771629 1.0 4.0     16.0
3.62715212  19.97137510 7.58503123  76.10193100 3.31749559  45.72032846 0.94747037  4.41042503  150597.85771629 1.0 4.0     16.0
3.16140219  8.09326275  1.95314020  4.58676418  118.06546646  58.80941584 0.21536658  4.29960013  148811.24340892 1.0 4.0   16.0
```
- Add both ILP and NEP potential files to **run.in** file and make sure the ILP file in the front:
```
potential ilp.txt nep.txt
```
- Make sure **element list** in NEP potential file is the **same** as ILP potential file, including the **ORDER**.

**Test**
- Energy conservation test in bilayer MoS2:
image
- `DEBUG` flag test:
image

**Others**
If any questions, please feel free to let me know. THANKS! :smiley_cat:

PR #951, opened 2025-03-21
Flexible version of hybrid potential NEP+ILP for vdW materials

BBBuZHIDAO:
## Summary
This hybrid potential is a flexible and optimized version of **NEP+ILP**, which can be used to simulate vdW materials, graphene, h-BN and TMDs now. (Water, metal interface and others will be added later) Not the same as the versions before (#835 #860 ), this version could read more NEP potential files to set the intralayer force field for each layer. You could decide which NEP potential file to use for the layer flexibly by a **map** file. The speed of this highly combined potential, with 16 angstrom ILP cutoff,  is in the order of **1e6** atom * step / second, tested in RTX4090.

## Modification
Add potential CUDA files (**ilp_nep.cu**  and **ilp_nep.cuh**) and check this potential in **force.cu** file.

## Usage
Three kinds of files are required: one ILP potential file, one map file for NEP and some NEP files.
- The format of ILP potential file is the same as #860 **except the potential name is not ilp_nep_tmd but ilp_nep**.
- The map file defines the correspondence between NEP potential files and layers. For example, the map file for trilayer graphene/graphene/MoS2 is:
```
<group_method_for_NEP>  2  C.nep  MoS.nep
3
0
0
1
```
  The first number is **group method** to distinguish the NEP layers, which means the atoms in each group of this group method will be set the same NEP force field. Usually, this group method is the same with the group method for ILP layers. The next number is your the number of your NEP potential files. And then add your NEP potential file paths behind. Here, two NEP potential files are added: C.nep (NEP 0) and MoS2.nep (NEP 1). 
The number in the second line is the number of layers in your structure model. Generally, this number is the number of groups in <group_method_for_NEP>. Finally, the last lines set the correspondence. In this example, the first 0 means group 0 uses C.nep (NEP 0), the second 0 means group 1 uses C.nep (NEP 0) and number 1 means group 2 uses MoS.nep (NEP 1).
- The NEP potential files are not required modification. But note that: **all atom types in NEP potential files need be in the single ILP potential file**

## Others
TODO: 
- [ ] `DEBUG` and `USE_TABLE` flags have been tested. More test will be added later.
- [ ] Update doc to show  more details and examples.

**If any questions, please feel free to let me know.**


Ali-SE98:
Hi, a quick question:
Is it possible to simulate any vdW 2D material with this last version of NEP+ILP (e.g. 2D MOFs and COFs), given that the ILP parameters are fitted for that specific 2D material?

BBBuZHIDAO:
> Hi, a quick question: Is it possible to simulate any vdW 2D material with this last version of NEP+ILP (e.g. 2D MOFs and COFs), given that the ILP parameters are fitted for that specific 2D material?

Hi, ILP parameters for 2D MOFs and COFs are not fitted now so it is a pity that nowadays ILP potential files can not achieve to simulate them. But this work is already going on! If the parameters is fitted, this NEP+ILP version could be used directly with the same normal vector definition as graphene and h-BN.

Ali-SE98:
> > Hi, a quick question: Is it possible to simulate any vdW 2D material with this last version of NEP+ILP (e.g. 2D MOFs and COFs), given that the ILP parameters are fitted for that specific 2D material?
> 
> Hi, ILP parameters for 2D MOFs and COFs are not fitted now so it is a pity that nowadays ILP potential files can not achieve to simulate them. But this work is already going on! If the parameters is fitted, this NEP+ILP version could be used directly with the same normal vector definition as graphene and h-BN.

Thanks.

Issue #473, opened 2023-08-15
Suggestion: Header for thermo.out file


erhart1:
It would be useful to include a header in the `thermo.out` file that specifies the meaning of the columns.
If these lines are prepended by a widely used comment symbols (e.g., `#`), the user experience should be (hopefully) unaffected.
At the same time it would help users (and parsers) to read the content of the file.
This would in particular improve readability/interpretability if the number of columns/data changes, e.g., when switching from an orthorhombic to a triclinic cell.
Finally, this would also be useful if one were to decides to add more information to this file in the future.

hityingph:
Yes, I agree with. Acutally, I also suggested this before, however, Zheyong @brucefan1983 prefered the pure arrays output without strings (header).

brucefan1983:
Perhaps we can do this for all the output files in GPUMD-v4.0, for which we will introduce breaking changes?

brucefan1983:
finished in PR #975

PR #975, opened 2025-04-12
add header to thermo.out

brucefan1983:
* To solve issue [#473 ](https://github.com/brucefan1983/GPUMD/issues/473)

* The followoing header will be added to `thermo.out`, for each `run` (so there can be multiple headers in one file):
```
# col 1: T K
# col 2: K eV
# col 3: U eV
# col 4: Pxx GPa
# col 5: Pyy GPa
# col 6: Pzz GPa
# col 7: Pyz GPa
# col 8: Pzx GPa
# col 9: Pxy GPa
# col 10: ax A
# col 11: ay A
# col 12: az A
# col 13: bx A
# col 14: by A
# col 15: bz A
# col 16: cx A
# col 17: cy A
# col 18: cz A
```

@elindgren Please check if this affects `calorine`.


erhart1:
> To solve issue [#473 ](https://github.com/brucefan1983/GPUMD/issues/473)
> 
> @elindgren Please check if this affects `calorine`.

`calorine` is not affected since it uses the [`loadtxt`](https://numpy.org/doc/2.2/reference/generated/numpy.loadtxt.html) function from numpy, which by default drops all rows that start with `#`.
I would consider using a standardized format for the column labels, so that they could in principle be parsed.
For example:
```
# col 1: [something] [unit]
# col 2: [something] [unit]
```

brucefan1983:
like this ?
image

erhart1:
> like this ? image

Although the brackets are not needed. I only meant to indicate generic field names

brucefan1983:
> > like this ? image
> 
> Although the brackets are not needed. I only meant to indicate generic field names

OK, I will remove the brackets.

brucefan1983:
note that for multiple runs, there will also be multiple headers.

PR #989, opened 2025-04-22
Fix bug: compilation failed with USE_TABLE for nep_ilp

BBBuZHIDAO:
**Summary**
`USE_TABLE` flag is not developed now, so should remove it in NEP+ILP potential (#951 ).

**Modification**
Comment out the code of `USE_TABLE` and remind in doc.

**Others**
`USE_TABLE` will be developed later.


BBBuZHIDAO:
Here is the test output by using NEP_ILP potential.
Compile without `USE_TABLE`:
image

Compile with `USE_TABLE`"
image

Issue #1007, opened 2025-05-02
Bug report about Plumed plugin


Ternity:
## Packages version
GPUMD v4.0 and Plumed v2.8.3
## Compiler environments
| c++ (for Plumed) | MPI (For Plumed) | CUDA
|---| ---| ---|
| Intel OneAPI compiler 2023.2.0 | Intel OneAPI MPI 2021.10.0 | CUDA toolkit 11.3 |

## Bug Report
When I try to compile GPUMD with plumed plugin, I get an error:
```bash
nvcc -std=c++11 -O3 -arch=sm_60 -DUSE_PLUMED -I./ -I/Path/to/Plumed/2.8.3/include -c measure/plumed.cu -o measure/plumed.o

measure/plumed.cu(99): error: incomplete class type "Integrate" is not allowed

measure/plumed.cu(188): error: class "Atom" has no member "force"

measure/plumed.cu(189): error: class "Atom" has no member "position"

measure/plumed.cu(201): error: class "Atom" has no member "virial"

measure/plumed.cu(221): error: incomplete class type "Force" is not allowed

measure/plumed.cu(236): error: class "Atom" has no member "virial"

measure/plumed.cu(237): error: class "Atom" has no member "virial"

measure/plumed.cu(238): error: class "Atom" has no member "virial"

measure/plumed.cu(239): error: class "Atom" has no member "virial"

measure/plumed.cu(240): error: class "Atom" has no member "virial"

measure/plumed.cu(241): error: class "Atom" has no member "virial"

11 errors detected in the compilation of "measure/plumed.cu".
```
### Error 1
For the first error `incomplete class type "Integrate" is not allowed`, it seems that the father class `Property` of class `Plumed` only provides the forward declaration of class `Integrate` during declaration:
https://github.com/brucefan1983/GPUMD/blob/b3f9be145da3b24b5a6df3cab1185bf6b5c1fcb2/src/measure/property.cuh#L24
But in the member functions of class `Plumed`, it was directly using its concrete method:
https://github.com/brucefan1983/GPUMD/blob/b3f9be145da3b24b5a6df3cab1185bf6b5c1fcb2/src/measure/plumed.cu#L99
### Error 2-11
For the other errors, the variables that do not exist in class `Atom` were used.
https://github.com/brucefan1983/GPUMD/blob/b3f9be145da3b24b5a6df3cab1185bf6b5c1fcb2/src/measure/plumed.cu#L188-L189
[...]

PR #1008, opened 2025-05-02
Fix compilation error of plumed plugin 

Ternity:
**Summary**
Aim to fix issue #1007 .

**Modification**
1. Add the head file path of class `Integrate` to `plumed.cuh` file.
2. Change the variables name of class `Atom`.

|before | force| position|virial|
|---|---|---|---|
|after|force_per_atom|position_per_atom|virial_per_atom|

**TODO**
More tests are needed to ensure consistent simulation results.


brucefan1983:
Good catch!

PR #1031, opened 2025-05-13
Added possibility to use simpler output file names for save_potential…

erhart1:
This is a convenience PR that adds a second parameter to the `save_potential` keyword in `nep.in` that allows the user to use a simpler output file name. The new format is `save_potential <number_of_generations_between_checkpoint_files> <format>`. If `<format>` is `0`, the checkpoint file name is `nep_gen<generation>.txt`. If `<format>` is `1` the current extended file name is used which includes the full date and time information. When writing in the latter format directories quickly fill up with long file names.

### DIFF
--- a/src/main_nep/parameters.cu
+++ b/src/main_nep/parameters.cu
@@ -100,7 +100,8 @@ void Parameters::set_default_parameters()
   use_full_batch = 0;          // default is not to enable effective full-batch
   population_size = 50;        // almost optimal
   maximum_generation = 100000; // a good starting point
-  save_potential = 100000;         // by default write a checkpoint nep.txt file every 100000 iterations
+  save_potential = 100000;     // write checkpoint nep.txt files at these intervals
+  save_potential_format = 1;   // 1 = include time stamp when writing checkpoint nep.txt files
   initial_para = 1.0f;
   sigma0 = 0.1f;
   atomic_v = 0;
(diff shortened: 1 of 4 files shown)


erhart1:
ping @brucefan1983 @elindgren

brucefan1983:
looks good to me.

src/main_nep/parameters.cu
```diff
@@ -1284,13 +1285,19 @@ void Parameters::parse_save_potential(const char** param, int num_param)
 {
   is_save_potential_set = true;
 
-  if (num_param != 2) {
-    PRINT_INPUT_ERROR("save_potential should have 1 parameter.\n");
+  if (num_param != 3) {
+    PRINT_INPUT_ERROR("save_potential should have 2 parameters.\n");
```
elindgren, line 1289: should we set `save_potential_format=1` by default, and not enforce it to be explicitly set as is currently done?

elindgren (approved):
LGTM, just had one small comment

src/main_nep/parameters.cu
brucefan1983: I have no strong opinion here as this is a new keyword that has not been widely used yet.

elindgren: Fair enough, lgtm then :+1:

PR #1041, opened 2025-05-16
TI-liquid sym bug fixed

tobias-hainer:
**Summary**

- Fixed a bug in the symmetrization of ti liquid. 
- Removed some comments used during development.
- Code now gives error when user specifies unimplemented p parameter.


brucefan1983:
thanks! Havee you finished?

tobias-hainer:
Yes, it is finished

PR #1045, opened 2025-05-19
Fix a bug.

psn417:
I fixed a bug in my code.


psn417:
This bug produced wrong result when calculating the extrapolation grade of a system with fewer types of elements than `nep.txt` (such as H2O for a H-O-Cl nep potential).

brucefan1983:
Good fix.

PR #1051, opened 2025-05-21
Msd many groups

elindgren:
**Summary**

Adds two options to `compute_msd`.
- `all_groups` which computes the `MSD` for each group individually in the selected grouping method.
- `save_every N` which writes a `msd_step[step].out` file every `N` MD steps. Useful for long calculations.

**Modification**

Added some logic in `msd.cu` for handling the new keywords, as well as a kernel function `gpu_find_msd_per_group`. The kernel function uses atomics for simplicity, and since atoms in the same group are not necessarily close in memory. Can probably be improved.

I've also added tests in the `/tests/gpumd/msd/` directory, which ensure the self-consistency of the `compute_msd` method. Additionally, I've checked that using `compute_msd` to compute the MSD for the full system yields the same result on master as on my branch, so this change should not have broken anything. 

**Others**

I've compared the implementation for computing the MSD over all groups in GPUMD (labelled GPUMD in the figure) to a naive python implementaton (Naive), an efficient approach using the wiener-kinchin theorem (FFT), as well as a reimplementation of the GPUMD algorithm (Python). The Naive-approach and the FFT agree perfectly, and the GPUMD implementation agrees with the reimplementation up to numerical precision of the `msd.out` file. The reason why the GPUMD implementation does not agree to numerical precision with the Naive/FFT implementations that use the definition of the MSD is because GPUMD does not use all frames of the trajectory when computing the MSD on the fly, I think.

image


brucefan1983:
nice! Many user wanted to have this. I will review asap.

doc/gpumd/input_parameters/compute_msd.rst
```diff
@@ -20,12 +20,28 @@ with parameters defined as
...
+The parameters are:
+
+* :attr:`all_groups <group_method>`, where :attr:`group_method` is the grouping method to use for computation
+
+Finally, the third special keyword is :attr:`save_every`.
+This keyword saves the internal :term:`MSD` and :term:`SDC` computed so far during the simulation, which can be helpful during long running simulations.
+The file will have a name formatted as ``msd_step[step].txt``. 
+The parameters are:
+
+* :attr:`save_every <interval> <group>`, where :attr:`interval` is the number of steps between saving a copy.
```
brucefan1983, line 42: `<group>` is redundant here

doc/gpumd/input_parameters/compute_msd.rst
```diff
@@ -39,3 +55,8 @@ This means that you
 * the position data will be recorded every 5 steps
 * the maximum number of correlation steps is 200
 * you would like to compute only over group 1 in group method 1.
+
+
+To compute the :term:`MSD` for all groups in group method 1 and save a copy of the :term:`MSD` every 100 steps, one can write::
```
brucefan1983, line 60: Is 100 a good number? Isn't it too frequent? What a good value?

doc/gpumd/input_parameters/compute_msd.rst
elindgren: good catch! forgot to remove it there

doc/gpumd/input_parameters/compute_msd.rst
elindgren: I would say that a good number is probably around 100 000-1 000 000 steps. shorter simulations are probably so quick that it's not needed to save the msd intermittently

brucefan1983: then how about modify the number in the example?

brucefan1983:
The figure above has unusaul units. Could you check it?

doc/gpumd/input_parameters/compute_msd.rst
elindgren: Will do!

brucefan1983:
The reason why the GPUMD implementation does not agree to numerical precision with the Naive/FFT implementations that use the definition of the MSD is because GPUMD does not use all frames of the trajectory when computing the MSD on the fly, I think.

-- I think so. The GPUMD implementation only starts when there are `Nc` frames accumulated.

elindgren:
> The figure above has unusaul units. Could you check it?

Do you mean my plot? It's just Ångström and femtoseconds, so no unusual units?

doc/gpumd/input_parameters/compute_msd.rst
```diff
@@ -20,12 +20,28 @@ with parameters defined as
...
+This keyword computes the :term:`MSD` and :term:`SDC` for each group in the specified grouping method.
+Note that :attr:`group` and :attr:`all_groups` cannot be used together.
+A typical usecase could be to compute the :term:`MSD` for each molecule in a system.
+The parameters are:
+
+* :attr:`all_groups <group_method>`, where :attr:`group_method` is the grouping method to use for computation
+
+Finally, the third special keyword is :attr:`save_every`.
+This keyword saves the internal :term:`MSD` and :term:`SDC` computed so far during the simulation, which can be helpful during long running simulations.
+The file will have a name formatted as ``msd_step[step].txt``.
```
brucefan1983, line 39: perhaps it is .out, not .txt？

elindgren:
> A correlation time of 0.01 fs? Even smaller than a time step?
> And MSD of 10^-6 A^2? Nearly non-diffusive or what?

Ah, I see what you mean. the unit on the x-axis should be ps, not fs. 
This is an extremely short run, it's just over 100fs where I dump every frame, with a correlation time of 20fs. so that is why the MSD is so small, the atoms haven't really had time to move. But I needed a short and quick run for comparison/debugging.

brucefan1983:
> > A correlation time of 0.01 fs? Even smaller than a time step?
> > And MSD of 10^-6 A^2? Nearly non-diffusive or what?
> 
> Ah, I see what you mean. the unit on the x-axis should be ps, not fs. This is an extremely short run, it's just over 100fs where I dump every frame, with a correlation time of 20fs. so that is why the MSD is so small, the atoms haven't really had time to move. But I needed a short and quick run for comparison/debugging.

OK, that sounds reasonable.

src/measure/msd.cu
```diff
@@ -130,7 +130,59 @@ __global__ void gpu_find_msd(
...
+        double dz = g_z[n] - g_z_start[size_sum + n];
+
+        if (bid <= correlation_step) {
+          int offset = group * num_correlation_steps + correlation_step - bid;
+          atomicAdd(&g_msd_x[offset], dx * dx);
+          atomicAdd(&g_msd_y[offset], dy * dy);
+          atomicAdd(&g_msd_z[offset], dz * dz);
+        } else {
+          int offset = group * num_correlation_steps + correlation_step + gridDim.x - bid;
+          atomicAdd(&g_msd_x[offset], dx * dx);
```
brucefan1983, line 178: this will not compile for Kepler GPU

elindgren: Nice catch! Fair enough about `atomicAdd`. Do you want me to change that before we merge this?

brucefan1983: depends on you. At least add that fix for Kepler.

PR #1085, opened 2025-06-25
Clean up msd many groups

### LINKED ISSUE #1055: migrate `compute_msd all_groups` from using `atomicAdd`
The function  `compute_msd all_groups` uses `atomicAdd`. `atomicAdd` is not necessarily an elegant nor efficient solution. We should instead use something more akin to a reduction.

elindgren:
Fixes #1055 

Changes from `atomicAdd` to using a reduction over groups to compute the MSD for each group.
Also clarifies a point regarding `save_every` in the documentation.

### DIFF
--- a/doc/gpumd/input_parameters/compute_msd.rst
+++ b/doc/gpumd/input_parameters/compute_msd.rst
@@ -39,7 +39,7 @@ This keyword saves the internal :term:`MSD` and :term:`SDC` computed so far duri
 The file will have a name formatted as ``msd_step[step].out``. 
 The parameters are:
 
-* :attr:`save_every <interval>`, where :attr:`interval` is the number of steps between saving a copy.
+* :attr:`save_every <interval>`, where :attr:`interval` is the number of steps between saving a copy. Note that the copy can only be written at most every :attr:`sample_interval` steps. Furthermore, the first ``msd_step[step].out`` file will be written after :attr:`Nc` times :attr:`sample_interval` steps. Subsequent files will be written every :attr:`interval`.
 
 
 Examples
--- a/src/measure/msd.cu
+++ b/src/measure/msd.cu
@@ -27,6 +27,7 @@ Calculate:
 #include "utilities/gpu_macro.cuh"
 #include "utilities/read_file.cuh"
 #include <cstring>
+#include <iostream>
 
 
 #ifdef USE_KEPLER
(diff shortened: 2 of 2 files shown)


elindgren:
@brucefan1983 This works, but I use a dynamically allocated array in shared memory to perform the reduction. This works for systems with modest number of groups, but it fails for systems with a lot of groups as the allocation becomes too large. The memory requirement is `3 * num_groups * block_size * sizeof(double)`. For example, I want to use it for a system with 2000 groups, which with a block size of 128 becomes `6144 kB`, which is significantly more than the 48 kB that is the maximum "safe" (i.e. what most GPUs have) of shared memory. The current setup could at maximum handle 15 groups before reaching the 48 kB limit. 

Do you have any suggestions for how to rewrite the kernel, such that we get rid of this memory issue? Ideally the kernel should be able to handle the edge case where every atom in the system has it's own group (which the old version using `atomicAdd` could, as long as the global GPU memory could fit the `msd` array).

elindgren:
ping @erhart1

brucefan1983:
have you compared speed? If the version with atomicAdd is fast enough, I do not insist on changing it to the new version.

elindgren:
> have you compared speed? If the version with atomicAdd is fast enough, I do not insist on changing it to the new version.

Did it know, and in kernel execution speed the `atomicAdd` version is actually faster, 9 us compared to 49 us. 
Note that I used a small test system of 480 atoms, corresponding to the maximum 15 molecules (groups) of the system I'm interested in. The atomicAdd version has easily handled the 1000 groups I've used before.  

In overall execution speed, the example takes 0.2934s to execute with the reduction, and 0.3102s using the master branch (`atomicAdd`). Overall, I think using `atomicAdd` should be fine.

Timing information for master version (`atomicAdd`) 

```bash
    Section: GPU Speed Of Light Throughput
    ----------------------- ----------- ------------
    Metric Name             Metric Unit Metric Value
    ----------------------- ----------- ------------
    DRAM Frequency                  Ghz         9.21
    SM Frequency                    Ghz         1.36
    Elapsed Cycles                cycle       12,803
    Memory Throughput                 %         7.86
    DRAM Throughput                   %         2.94
    Duration                         us         9.41
    L1/TEX Cache Throughput           %        21.59
    L2 Cache Throughput               %         7.86
    SM Active Cycles              cycle     2,132.76
    Compute (SM) Throughput           %         2.88
    ----------------------- ----------- ------------
```

Timing information for this version (`reduction`) 
```bash
    Section: GPU Speed Of Light Throughput
    ----------------------- ----------- ------------
    Metric Name             Metric Unit Metric Value
    ----------------------- ----------- ------------
    DRAM Frequency                  Ghz         9.23
    SM Frequency                    Ghz         1.36
    Elapsed Cycles                cycle       68,119
    Memory Throughput                 %         1.44
    DRAM Throughput                   %         0.56
    Duration                         us        49.95
    L1/TEX Cache Throughput           %         6.01
    L2 Cache Throughput               %         0.71
    SM Active Cycles              cycle    16,247.58
    Compute (SM) Throughput           %         2.91
    ----------------------- ----------- ------------
```

elindgren:
I profiled the full 32000 atom system (1000 groups) with `atomicAdd` for fun, and the kernel only took 250 us to execute. So I would hazard and say that computing the MSD is not a major cost anyways.

```bash
 ----------------------- ----------- ------------
    Metric Name             Metric Unit Metric Value
    ----------------------- ----------- ------------
    DRAM Frequency                  Ghz         9.24
    SM Frequency                    Ghz         1.36
    Elapsed Cycles                cycle      376,503
    Memory Throughput                 %        17.21
    DRAM Throughput                   %         7.46
    Duration                         us       275.87
    L1/TEX Cache Throughput           %        33.16
    L2 Cache Throughput               %        17.21
    SM Active Cycles              cycle    90,514.29
    Compute (SM) Throughput           %         6.50
    ----------------------- ----------- ------------ 
```

elindgren:
In general the profiling reports indicate that we are leaving quite a bit of performance on the table. For example, my kernel only achieves an occupancy of 8.3%. This is something I've noticed in other places when profiling the code. @brucefan1983 do you think we should take a look at systematically profile and optimize key parts of the code (e.g. force evaluations)? if so, how do we optimize for all different target gpus that users might have?

elindgren:
I've cleaned up and reverted to using `atomicAdd`. @brucefan1983 feel free to merge this if you're happy.

brucefan1983:
OK, let's keep the old version. 

For force evaluation, I think there is not much to do for now.

PR #1087, opened 2025-06-27
More compute msd bugs

elindgren:
Fixes two bugs pointed out by @erhart1. 

- 1. msd will be `-nan` for empty groups
- 2. msd and sdc will be `-nan` when the msd correlation length is longer than the md run length in steps.

Both bugs comes from the `msd_scaler`, which is `1 / (num_atoms * num_time_origins)`. In bug 1 `num_atoms=0`, and in bug 2 `num_time_origins=0`. 

In both cases, the `msd_scaler` variable is now set to 0, leading to the msd and sdc to be 0 instead of nan.

### DIFF
--- a/src/measure/msd.cu
+++ b/src/measure/msd.cu
@@ -367,8 +367,14 @@ void MSD::write(const char* filename)
   // normalize by the number of atoms and number of time origins
   for (int group_id=0; group_id < num_groups_; group_id++) {
     int num_atoms = num_atoms_per_group_[group_id];
-    // num_time_origins_ should be different for each nc
-    const double msd_scaler = 1.0 / ((double)num_atoms * (double)num_time_origins_);
+    
+    // This is the case for empty groups
+    double msd_scaler = 0.0;
+    if (num_atoms > 0) {
+      // num_time_origins_ should be different for each nc
+      msd_scaler = 1.0 / ((double)num_atoms * (double)num_time_origins_);
+    } 
+
     int group_index = group_id * num_correlation_steps_;
 
     for (int nc = group_index + 0; nc < group_index + num_correlation_steps_; nc++) {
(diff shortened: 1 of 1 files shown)


elindgren:
ping @erhart1

brucefan1983:
I would say both are input errors and a better way is to report error mesages and exit.

elindgren:
> I would say both are input errors and a better way is to report error mesages and exit.

I can agree that bug 2 could be considered an input error, no MSD will have been computed anyway. But 1. could potentially occur when working with a lot of different groups over many runs. Maybe @erhart1 could fill in with some details on his particular usecase?

erhart1:
There are two "main" problems:

1. The error is not visible until the completion of a (possibly very long) run (even when using the intermediate save functionality).
2. The output is `-nan` and that the columns are not separated. Hence the output will contain fewer columns than expected. For example, assuming two groups with the first one being empty
  ```shell
  100-nan-nan-nan-nan-nan 10 20 30 40 50 60
  ```
One can fix this problem after the fact, e.g., via `sed -i 's/-nan/ nan/' msd.out`, but that is not optimal from a usability stand point.

One could raise an error if a division by zero is attempted. Or use a standardized output value. In the latter case, I would prefer `-1` (or `-nan` with a space ;-)) to clearly indicate the non-physicality of the output.

brucefan1983:
I wonder why there is empty group.

erhart1:
Fair question. It can happen when setting up runs over many compositions ;-)

brucefan1983:
how about we report error message for the second nan, and keep the present solution for the first nan?

elindgren:
> how about we report error message for the second nan, and keep the present solution for the first nan?

I think that makes sense.

PR #1093, opened 2025-07-02
Development of GNEP Based on NEP: Introducing Full-Parameter Gradients, Adam Optimizer, and Cosine Annealing Training

hfood02:
**Summary**

Based on the original NEP (Neuroevolution Potential), a new method called Gradient-optimized Neuroevolution Potential (GNEP) has been developed. This approach introduces full-parameter gradient computation, integrates the Adam optimizer, and employs cosine annealing for efficient potential training.

**Modification**

Added core GNEP-related code, including parameter gradient computation modules.
Integrated the Adam optimizer for adaptive learning rate adjustment.
Introduced cosine annealing learning rate scheduling to improve training convergence and stability.
Extended the input file format (`gnep.in`) to support GNEP-specific parameter configuration.

**Others**

Compatible with the original NEP/GPUMD framework, supporting one-click compilation and execution.
Currently supports single-GPU training; multi-GPU support can be added in the future.
For theoretical background and implementation details, refer to arXiv:2507.00528.

### DIFF
--- a/src/main_gnep/gnep.cu
+++ b/src/main_gnep/gnep.cu
@@ -655,19 +655,17 @@ static __global__ void compute_grad_radial_NM(
   float* e_wb_grad = g_E_wb_grad + n1_net_index;
   float grad_w0_sum, grad_w1_sum, grad_b0_sum;
   if (no_neighbor_isolated_atom) {
-    for (int j = 0; j < annmb.num_neurons1; ++j) {
-      for (int d = 0; d < annmb.dim; ++d) {
-        grad_w0_sum += i1 == 0 ? per_Nc_e * e_wb_grad[j * annmb.dim + d] : 0.0f;
-        atomicAdd(&g_grad_sum[t1_net_index + j * annmb.dim + d], grad_w0_sum);
-      }
-      if (i1 == 0) {
-        grad_w1_sum += e_wb_grad[b0_index + j] * per_Nc_e;
-        grad_b0_sum += e_wb_grad[w0_index + j] * per_Nc_e;
-      }
-      atomicAdd(&g_grad_sum[t1_net_index + b0_index + j], grad_w1_sum);
-      atomicAdd(&g_grad_sum[t1_net_index + w0_index + j], grad_b0_sum);
-    }
     if (i1 == 0) {
+      for (int j = 0; j < annmb.num_neurons1; ++j) {
+        for (int d = 0; d < annmb.dim; ++d) {
+          grad_w0_sum = per_Nc_e * e_wb_grad[j * annmb.dim + d];
+          atomicAdd(&g_grad_sum[t1_net_index + j * annmb.dim + d], grad_w0_sum);
+        }
+        grad_w1_sum = e_wb_grad[b0_index + j] * per_Nc_e;
+        grad_b0_sum = e_wb_grad[w0_index + j] * per_Nc_e;
+        atomicAdd(&g_grad_sum[t1_net_index + b0_index + j], grad_w1_sum);
+        atomicAdd(&g_grad_sum[t1_net_index + w0_index + j], grad_b0_sum);
+      }
       float e_b1_n1 = e_wb_grad[annmb.num_neurons1 * annmb.dim + annmb.num_neurons1 + annmb.num_neurons1] * per_Nc_e;
       atomicAdd(&g_grad_sum[t1_net_index + annmb.num_neurons1 * annmb.dim + annmb.num_neurons1 + annmb.num_neurons1], e_b1_n1);
     }


elindgren:
Interesting, but I have a few questions.

*  As you state in [your paper](https://arxiv.org/pdf/2507.00528), the main advantage of SNES is increased exploration of the parameter space, but at a higher cost. I would thus expect NEP to be less prone to overfitting than GNEP. Do you have plots of the **test error** as  a function of training epoch for both methods, similar to Figure 1 in the paper? 
* How interoperable with NEP is GNEP? Could one initially train for a certain number of generations with SNES, and then continue with GNEP from the `restart.txt` file? Looking at the loss curves in Figure 1 of the paper, that seems like the most efficient and would combine 
* How are the L1 and L2 losses affected in GNEP compared to NEP? Strong regularization is a big selling point of NEP imo.
* What motivated the change of the loss function from using the RMSE to the MSE? I prefer the MSE, but I want to hear your reason :-) 

FIgure 1:
image

hfood02:
@elindgren 

Thanks for your interest and thoughtful questions!
1. In my current study, I did not observe clear signs of overfitting during GNEP training. The test errors remained stable or improved consistently across epochs, and validation RMSE curves closely tracked the training loss, suggesting that generalization was preserved throughout.
image

2. That's a very insightful suggestion—and in fact, I've tested this idea. Technically, it is indeed feasible. However, in my tests, I found that this hybrid approach does not yield the expected benefit. There are some main reasons:
    - SNES operates in a full-batch fashion with stochastic sampling over the entire parameter space. GNEP, in contrast, is built on mini-batch stochastic gradient descent (Adam), which assumes local curvature structure and benefits from early momentum accumulation.
    - Transitioning from SNES to Adam mid-training introduces inconsistency in gradient scale, momentum, and batch statistics, which often causes inefficient re-adaptation and slows convergence.
    - Starting from SNES checkpoints did not significantly reduce the required training steps, and in some cases even delayed convergence, as Adam had to “re-orient” itself from an inherited suboptimal parameter direction.
3. You're absolutely right that regularization plays a key role in controlling model generalization. In our current implementation: 
    - L2 regularization (weight decay) is naturally supported via [AdamW](https://arxiv.org/abs/1711.05101), which decouples weight decay from the gradient update. In my test, I didn't encounter any serious overfitting, so I didn't retain too much of this test.
    - L1 regularization, while possible to implement manually, is not easily realized for now. Going forward, integrating L1-like sparsity could be interesting.
4. Actually, I use the MSE for training, because its gradient is linear. However, I report RMSE during evaluation, since it shares the same unit as the physical quantity (e.g., meV/atom or meV/Å).

elindgren:
> @elindgren
> 
> Thanks for your interest and thoughtful questions!
> 
>     1. In my current study, I did not observe clear signs of overfitting during GNEP training. The test errors remained stable or improved consistently across epochs, and validation RMSE curves closely tracked the training loss, suggesting that generalization was preserved throughout.
>        image
> 
>     2. That's a very insightful suggestion—and in fact, I've tested this idea. Technically, it is indeed feasible. However, in my tests, I found that this hybrid approach does not yield the expected benefit. There are some main reasons:
>        
>        * SNES operates in a full-batch fashion with stochastic sampling over the entire parameter space. GNEP, in contrast, is built on mini-batch stochastic gradient descent (Adam), which assumes local curvature structure and benefits from early momentum accumulation.
>        * Transitioning from SNES to Adam mid-training introduces inconsistency in gradient scale, momentum, and batch statistics, which often causes inefficient re-adaptation and slows convergence.
>        * Starting from SNES checkpoints did not significantly reduce the required training steps, and in some cases even delayed convergence, as Adam had to “re-orient” itself from an inherited suboptimal parameter direction.
> 
>     3. You're absolutely right that regularization plays a key role in controlling model generalization. In our current implementation:
>        
>        * L2 regularization (weight decay) is naturally supported via [AdamW](https://arxiv.org/abs/1711.05101), which decouples weight decay from the gradient update. In my test, I didn't encounter any serious overfitting, so I didn't retain too much of this test.
>        * L1 regularization, while possible to implement manually, is not easily realized for now. Going forward, integrating L1-like sparsity could be interesting.
> 
>     4. Actually, I use the MSE for training, because its gradient is linear. However, I report RMSE during evaluation, since it shares the same unit as the physical quantity (e.g., meV/atom or meV/Å).

Cool stuff, thanks for clarifying!
1. That's reassuring, thanks for testing! :+1: 
2. Interesting! I didn't know that Adam would require such slow re-adaptation.
3. Fair enough. In the long term I suppose L1 would be nice, using a combined loss (L1 + L2 = elastic net loss) is in my experience one of the nice with NEP, as the final models become rather sparse.
4. Good point that the loss becomes linear. I was mainly coming from a statistics perspective, where the interpretation of a MSE is much more intuitive than a RMSE.

brucefan1983:
I will review carefully when I have time. For the time being, I will keep it as draft. This is to prevent accidental merge.

src/force/force.cu
```diff
@@ -976,4 +976,4 @@ void Force::compute(
       GPU_CHECK_KERNEL
     }
   }
-}
\ No newline at end of file
+}
```
brucefan1983, line 979: This file could be completely restored.

src/main_gnep/adam.cu
```diff
@@ -0,0 +1,418 @@
+/*
```
brucefan1983, line 1: Copyright part should follow the other files. You can write your contribution information after the copyright part:

```
/*
    Copyright 2017 Zheyong Fan and GPUMD development team
    This file is part of GPUMD.
    GPUMD is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.
    GPUMD is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.
    You should have received a copy of the GNU General Public License
    along with GPUMD.  If not, see <http://www.gnu.org/licenses/>.
*/

/*------------------------------------------------------------------------------
Your contribution statement here.
------------------------------------------------------------------------------*/
```

This comment applies to all the similar  files.

src/main_gnep/adam.cu
```diff
@@ -0,0 +1,418 @@
...
+    eps,
+    weight_decay,
+    gpu_m.data(),
+    gpu_v.data(),
+    gpu_parameters.data()
+  );
+  GPU_CHECK_KERNEL
+
+  gpu_parameters.copy_to_host(parameters.data());
+  // printf("Parameters: \n");
```
brucefan1983, line 397: Do not keep commented code.

src/main_gnep/fitness.cu
```diff
@@ -0,0 +1,696 @@
...
+  : optimizer(adam)
+{
+  maximum_epochs = para.epoch;
+  number_of_variables = para.number_of_variables;
+  number_of_variables_ann = para.number_of_variables_ann;
+  number_of_variables_descriptor = para.number_of_variables_descriptor;
+  lr = para.lr;
+  start_lr = para.start_lr;
+  stop_lr = para.stop_lr;
+  // decay_step = para.decay_step;
```
brucefan1983, line 45: Do not keep commented code.

brucefan1983:
General comment: It is not good to keep commented code. I suggest you clean up carefully.

hfood02:
Ok. I'll clear out the redundant comments

PR #1106, opened 2025-07-29
Improving NEP-Charge

brucefan1983:
**Summary**

Improving NEP-Charge

**Modification**

* Confine NEP-Charge to the NEP4 verison. It is meaningless to use the NEP3 version for charge.
* Better initialization to speed up the training. The initial charge is zero for every ion.
* Can take Born effective charge (BEC) as a training target.

**TODOs**

* Need to deal with non-periodic systems.

* Should we implement PPPM, PME, or RBE in MD part?


luowh35:
For higher accuracy, I think PME is a better choice. This article proposes a method https://doi.org/10.1063/5.0251713, which the computational complexity is (O^3/2).

Issue #385, opened 2023-02-28
Feature requested: local order parameters


ZKC19940412:
As GPUMD massively upgrades its features, it will be nice to enable an on-the-fly computation of local order parameters. In essential, a local order parameter is used to identify the presence of different ice phases when one uses GPUMD to perform MD simulations on water/ice systems. Local order parameters can benefit various applications, such as probing ice nucleations. The functional form of order parameters mimics the spherical harmonics functions (more can be found in equations 1 and 2 in the reference paper), which are computed as the angular descriptors for NEP potential. Therefore, it should be very straightforward to may be added in a new "compute" keyword or include it in the  "dump" keyword. 

Please consider implementing this feature, and a reference paper can be found here: https://pubs.rsc.org/en/content/articlelanding/2011/cp/c1cp22167a

Thank you all so much.

Best, 
Zekun

brucefan1983:
Thanks for this good suggestion. We can consider enabling it in the near future.

brucefan1983:
I have started looking at the literature. It seems there is a more general set of quantities as proposed in this paper: 

[Bond-orientational order in liquids and glasses](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.28.784)

which is implemented in LAMMPS: https://docs.lammps.org/compute_orientorder_atom.html

Do you think this is a better choice? @ZKC19940412

brucefan1983:
We aim to achieve this in GPUMD-v3.9.

ZKC19940412:
Hi Bruce. Sorry for the late response. Yes. This reference in LAMMPS should be a good starting point. Thanks for working on it.

brucefan1983:
@ZKC19940412 Thanks for confirming this. I have just released GPUMD-v3.8, and I will aim to add this feature to GPUMD-v3.9.

brucefan1983:
This has been resolved by PR #1134. Big thanks to Dr. Yongchao Wu!

PR #1134, opened 2025-08-26
Add Steinhardt Bond-Orientational Order Parameter Computation

mushroomfire:
**Summary**  
This PR introduces the computation of Steinhardt bond-orientational order parameters, addressing issue #385.  
Currently, both *cutoff* and *nearest-neighbor number (nnn)* modes are supported to calculate these parameters for multiple degrees at once.  
A more detailed description and documentation will be uploaded later.

**Validation**  
- All results from [REF](https://arxiv.org/pdf/1209.6180) can be successfully reproduced.  
- Additionally, the method has been tested on a multielement alloy system simulated for 100 steps in an NPT ensemble. Results at step 100 were compared against values computed by **MDAPY**.  

Example input:  
```bash
compute_orientorder 100 nnn 12 3 4 6 8 1 1 1
```
Command output:
image

Comparison results:
image

**Others**
In a benchmark with 200,000 atoms, enabling this feature only introduced a negligible runtime overhead, demonstrating the efficiency of the implementation.


brucefan1983:
superb!

perhaps the screen output can be improved a little:

1) explain the number `100`
2) `dgrees` means angular number `l`? What is the official name?

mushroomfire:
I have modified the command output and updated the detailed documentation (tested locally). The new output is shown below:
image

Issue #1146, opened 2025-09-11
Integer overflow in ensemble_ti_spring parameter validation causes false error


Jonsnow-willow:
## Bug Description

The `Ensemble_TI_Spring` constructor incorrectly reports an error when valid `tswitch` and `tequil` parameters are provided due to integer overflow in the validation check.

## Steps to Reproduce

1. Use the following input parameters:
```
ensemble ti_spring temp 500 tperiod 100 tequil 40000 tswitch 100000 spring Ti 2.0
```

2. The program fails with error:
```
Error: Please specify either both t_switch and t_equil, or neither (to let the program auto-determine)
```

## Expected Behavior

The parameters should be accepted since both `tswitch` and `tequil` are properly specified with positive values.

## Actual Behavior

The program throws an error due to integer overflow in the validation check.

## Debug Output

```
DBG: t_equil=40000, t_switch=100000
DBG: t_equil * t_switch=-294967296
```

## Root Cause

In `ensemble_ti_spring.cu` line 143:
```cpp
if (t_switch * t_equil < 0) {
```

The multiplication `40000 × 100000 = 4,000,000,000` exceeds the maximum value for 32-bit signed integers (`2,147,483,647`), causing integer overflow and resulting in a negative value (`-294967296`).

## Proposed Solution

 Separate validation**
```cpp
// Check each parameter individually
if ((t_switch < 0 && t_equil >= 0) || (t_switch >= 0 && t_equil < 0)) {
    PRINT_INPUT_ERROR("Error: Please specify either both t_switch and t_equil, or neither...");
}
```

brucefan1983:
good catch. Do you want to fix it?

Jonsnow-willow:
Sure, I’ll fix it.

brucefan1983:
fixed in PR #1148

PR #1148, opened 2025-09-12
Refactor condition for t_switch and t_equil validation

Jonsnow-willow:
Resolved an integer overflow bug in ensemble_ti_spring parameter validation that could trigger false error messages.

Changes
	•	Corrected integer handling to avoid overflow.
	•	Ensured proper validation of tswitch and related parameters.


brucefan1983:
fixing issue #1146

Issue #1140, opened 2025-09-07
"minimize" keyword paired with the FIRE method failed to optimize small structures


astkr:
It is observed that the "minimize" keyword may fail to optimize small structures like the primitive cell of silicon which contains two atoms. To verify this, an [NEP model for silicon](https://github.com/brucefan1983/GPUMD-Tutorials/blob/main/potentials/nep/Si_2022_NEP3_4body.txt) proposed in 2022 is used in combination with ASE/calorine and GPUMD to perform structure optimization for a silicon primitive cell under `fmax = 1e-5` and `steps = 1e3`.

| Method              | Atoms              | Original lattice constant | Optimized lattice constant |
| ------------------- | ------------------ | ------------------------- | -------------------------- |
| ASE/BFGS            | primitive cell     | 5.434                     | 5.455                      |
| ASE/BFGSLineSearch  | primitive cell     | 5.434                     | 5.455                      |
| ASE/FIRE            | primitive cell     | 5.434                     | 5.455                      |
| ASE/FIRE2           | primitive cell     | 5.434                     | 5.455                      |
| ASE/GPMin           | primitive cell     | 5.434                     | 5.455                      |
| ASE/LBFGS           | primitive cell     | 5.434                     | 5.455                      |
| ASE/LBFGSLineSearch | primitive cell     | 5.434                     | 5.455                      |
| ASE/MDMin           | primitive cell     | 5.434                     | 5.455                      |
| ASE/SciPyFminBFGS   | primitive cell     | 5.434                     | 5.455                      |
| ASE/SciPyFminCG     | primitive cell     | 5.434                     | 5.455                      |
| GPUMD/FIRE          | primitive cell     | 5.434                     | The box inflates greatly   |
| GPUMD/FIRE          | super cell 2\*2\*2 | 5.434                     | The box inflates greatly   |
| GPUMD/FIRE          | super cell 4\*4\*4 | 5.434                     | The box inflates greatly   |
| GPUMD/FIRE          | super cell 8\*8\*8 | 5.434                     | 5.455                      |

From the results above, it seems that ASE's optimizers could optimize the primitive cell properly to the precision of `fmax = 1e-5`. However, GPUMD's implementation of the FIRE method cannot optimize small structures whose box might overly inflate during the process. The optimization of a 8\*8\*8 super cell yields the same lattice constant, 5.455 A, as ASE does.

[...]

PR #1164, opened 2025-09-21
Fix small cell minimization 

mushroomfire:
This PR resolves issue #1140 by enabling successful minimization of small cells. In addition to the Si example provided by the user, I have also tested the implementation with the primitive cells of aluminum and copper.


brucefan1983:
Good fix.

PR #1199, opened 2025-10-09
Add deposit command — atom deposition from a free surface

Jonsnow-willow:
**Summary**

Introduce a new command deposit to periodically add atoms from a specified region (typically above a free surface) into the simulation box.

This is a prototype implementation focused on basic functionality —
no neighbor list rebuild, no potential recalculation — aiming to test data-structure integrity and workflow stability.

**Purpose**
	•	Simulate atom deposition / surface growth inside GPUMD
	•	Test host–device synchronization when dynamically increasing atom count
	•	Lay groundwork for future sputtering / implantation simulations

**Modification**
    *Added*
	•	deposit.cu
	•	deposit.cuh
     *Command Intro*
        deposit <element>  x <x1> <x2> y <y1> <y2> z <z1> <z2> v <vx> <vy> <vz> <interval> [group <group_method> <group_label>]
       This command periodically deposits one atom within a specified (x, y, z) range, assigns a fixed velocity, and updates all host/device arrays accordingly. Currently, it does not include neighbor-list or potential recalculation — focusing on the minimal working prototype for surface deposition.
     *Example*
       deposit Cu x 0 20 y 0 20 z 30 35 v 0 0 -0.05 100

**Others**

This PR is an initial draft for discussing logic and field naming (e.g., symbol_, interval_, velocity_input_, velocity_natural_), and the complete implementation is planned within one week.

Although CUDA-level atom insertion ensures internal consistency, it adds coding complexity with limited performance gain. In most deposition cases, the time interval between atom insertions is long enough that handling such operations in Python would not significantly impact overall simulation speed. As a suggestion, we may consider developing a lightweight Python toolkit dedicated to GPUMD for structure generation and high-level automation. Of course, this is only a proposal — the final design direction should fully respect the your original intentions.

### DIFF
--- a/src/force/force.cu
+++ b/src/force/force.cu
@@ -487,6 +487,11 @@ void Force::compute(
   GPU_Vector<double>& virial_per_atom)
 {
   const int number_of_atoms = type.size();
+
+  for (auto& potential : potentials) {
+    potential->N1 = 0;
+    potential->N2 = number_of_atoms;
+  }
   if (!is_fcp) {
     gpu_apply_pbc<<<(number_of_atoms - 1) / 128 + 1, 128>>>(
       number_of_atoms,


Jonsnow-willow:
**update**

Currently, atom insertion has been implemented successfully.
However, the initial structure must still contain the species intended for later deposition, and the newly inserted atoms are not yet properly included in interatomic interaction calculations.

brucefan1983:
> **update**
> 
> Currently, atom insertion has been implemented successfully. However, the initial structure must still contain the species intended for later deposition, and the newly inserted atoms are not yet properly included in interatomic interaction calculations.

Is there a problem here? I think hte potential can be the one that includes all the involved species.

Jonsnow-willow:
I’ve already done
for (auto& potential : potentials) {
  potential->N1 = 0;
  potential->N2 = number_of_atoms;
}
but I’m not sure why the newly inserted atoms are still not included in the energy, force, and virial calculations.

brucefan1983:
> I’ve already done for (auto& potential : potentials) { potential->N1 = 0; potential->N2 = number_of_atoms; } but I’m not sure why the newly inserted atoms are still not included in the energy, force, and virial calculations.

I have not studied the details yet. My first impression is that one should initially set the number of atoms large enough, allocating enough memory for the potential class. Later use the value of N2 to control the range of calculation.

Jonsnow-willow:
Maybe we can treat the atoms that need to be deposited as pseudo-atoms, with their group ID set to –1, so they are ignored during the calculations. The deposit command would then change their group to a positive value and assign them a velocity.

This approach would require modifying the current ‘group’ design, so I’d like to discuss it with you before proceeding. However, it has a significant advantage — after this change, we could extend the implementation to support atom addition or deletion operations in VCSGC (e.g., for interstitial atoms).

Jonsnow-willow:
After some experiments, we believe that building a standalone Python package — independent of external software — provides a more flexible and efficient solution for handling atom deposition. Therefore, we are closing this PR.

PR #1214, opened 2025-10-24
publications bib add

duanzaixu:
add


duanzaixu:
add
Strain-induced conformation transition in two-dimensional covalent organic frameworks
and other

Issue #1315, opened 2026-01-23
compute_rdf extension


brucefan1983:
Present version has an upper bound on the number of atom pairs; need an extension.

brucefan1983:
solved in PR #1336

PR #1336, opened 2026-02-03
RDF generalization and speedup

brucefan1983:
**Summary**
Solves issue #1315 

* Usage of `compute_rdf` is simplified. There is no need to specify the atom pairs any more and the code will calculate the partial RDFs for all atom pairs.

* The calculation is significantly sped up and the memory usage is significantly reduced.

* Eaxample usage:
```
compute_rdf 6 200 20
``` 

* Screen output 
```
Compute radial distribution function (RDF).
    radial cutoff 6.
    radial cutoff will be divided into 200 bins.
    RDF sample interval is 20 step.
    There are 2 atom types in model.xyz.
        Type 0 has 512 atoms.
        Type 1 has 1024 atoms.
    Will calculate one total RDF and 3 partial RDFs.
```

Results are not changed:
image

### DIFF
--- a/doc/gpumd/input_parameters/compute_rdf.rst
+++ b/doc/gpumd/input_parameters/compute_rdf.rst
@@ -33,18 +33,13 @@ Syntax
 
 This keyword is used as follows::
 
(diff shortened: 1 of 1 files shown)


yongwangxx:
Confirmed

PR #1354, opened 2026-02-13
Neighbor skin

brucefan1983:
Many potentials are sped up.


brucefan1983:
> I have reviewed all modified code. It's a very clever strategy by using skin to update the NL! I also test the hybrid ILP potentials. These modifications are safe for ILPs. The only suggestion is that if it's necessary to set the `skin` as a macro definition (maybe some simulations need high atomic speed, then if the skin is not sufficient, global NL will still be updated as last versions).

Using a larger skin will also make the construction for the global and local neighbor lists slower. I can test the optimal skin distance in the future.

PR #1389, opened 2026-03-08
Modify the processing method prior to phonon dispersion calculation

tang070205:
Modifications to Pre-Phonon Dispersion Calculation Processing
1. Remove the `basis.in` . Instead, directly read the relative atomic mass of the corresponding atom from the atom class and construct content identical to that in the `basis.in` .

2. Use the `replicate` parameter from the run.in file to construct the supercell.

3. To change the `kpoints.in` file format, simply input the high symmetry points. As shown below
`0 0 0 G`
`0.5 0 0 M`
`0.33333 0.33333 0 K`
`0 0 0 G`
or
`0 0 0 G`
`0.5 0 0.5 X`
`0.625 0.25 0.625 U`
`        `  #Represents a breakpoint, marking the start of the second path.
`0.375 0.375 0.75 K`
`0 0 0 G`
`0.5 0.5 0.5 L`
`0.5 0.25 0.75 W`
`0.5 0 0.5 X`

4. The output format for omega2.out has changed. The first line now outputs the q-point and its corresponding wave vector value. A new first column has been added below, listing the wave vector values for each k-point. As shown in the figure below
image

5. For `run.in` file, the replicate parameter must be specified.
An uncertain value can be used, and a prompt will be displayed during subsequent calculations.
image

compute_phonon does not require manual input of truncation; the program internally determines this automatically. Therefore, only displacement data needs to be entered.
`compute_phonon 12 0.01`  #Assuming a radial truncation of 6 Å
change to
`compute_phonon 0.01`

In previous usage, it required a large amount of complex input, though this could be generated via scripts, such as the following script:
[https://github.com/tang070205/tools/blob/main/create_phonon_compare.py](url)
image
As can be seen, we need to know a large number of parameters. The `basis.in` and `kpoints.in` file format remains unchanged from before. The contents of `run.in` are as follows:
`potential nep.txt`
`compute_phonon  12 0.01`
The output files also contain the square of the angular frequency.
image

The revised usage method eliminates complex preprocessing. The `model.xyz` file now requires only primitive cells, while high-symmetry points are directly specified in the `kpoints.in` file.
`0 0 0 G`
`0.5 0 0.5 X`
`0.625 0.25 0.625 U`
`        ` 
`0.375 0.375 0.75 K`
`0 0 0 G`
`0.5 0.5 0.5 L`
`0.5 0.25 0.75 W`
`0.5 0 0.5 X`

After modification, the `run.in` file must use the `replicate` parameter. However, there is no need to set a truncation value anymore.
`replicate 5 5 5
`potential nep.txt`
`compute_phonon  0.01`

The output file now includes the high symmetry points and wave vector positions in the first row, along with the corresponding wave vector values for each k-point in the first column.
image

phonon

[test-UO2.zip](https://github.com/user-attachments/files/25859694/test-UO2.zip)

### DIFF
--- a/doc/gpumd/input_files/kpoints_in.rst
+++ b/doc/gpumd/input_files/kpoints_in.rst
@@ -14,8 +14,8 @@ The format of this file must be as follows:
 
 .. code::
 # high symmetry points
-points_x(0) points_y(0) points_z(0) name(0)
-points_x(1) points_y(1) points_z(1) name(1)
+k_points_x(0) k_points_y(0) k_points_z(0) name(0)
+k_points_x(1) k_points_y(1) k_points_z(1) name(1)
 ...
 
 
@@ -46,5 +46,5 @@ or
  0.5 0 0.5 X
 
 Here,
-* The first three columns represent the coordinates of high symmetry points.The last column corresponds to names.
+* The first three columns represent the coordinates of high symmetry points. The last column corresponds to names.
 * A blank line indicates a breakpoint or the initiation of a second path.
--- a/doc/gpumd/input_parameters/compute_phonon.rst
+++ b/doc/gpumd/input_parameters/compute_phonon.rst
@@ -18,7 +18,7 @@ To use this keyword, please make sure the following files have been preprared in
    * - Input filename
      - Brief description
    * - ``kpoints.in``
-     - High Symmetric Point Path, coordinates and names
+     - Specify the :math:`\boldsymbol{k}`-points along the high-symmetry paths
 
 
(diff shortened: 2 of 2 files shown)


brucefan1983:
Great improvements!
 
General comments: 
* Make comprehensive tests, comparing with old code. 
* Only keep the new documentation; don't need to keep those marked as "before version 5.0".

tang070205:
> Great improvements!
> 
> General comments:
> 
> * Make comprehensive tests, comparing with old code.
> * Only keep the new documentation; don't need to keep those marked as "before version 5.0".

I added an example for testing, and multiple paths still work fine.
The manual has also been updated to the new version.

PR #1428, opened 2026-04-03
Add a second hidden layer for testing

tang070205:
Change the second parameter of the ANN in nep.txt from 0 to the number of neurons in the second hidden layer


src/main_nep/parameters.cu
```diff
@@ -364,13 +369,21 @@ void Parameters::check_foundation_model()
 
   // 7th line, ANN
   tokens = get_tokens(input);
-  if (tokens.size() != 3) {
+  if (tokens.size() < 2 || tokens.size() > 3) {
```
brucefan1983: don't need to modify this part, as this is for NEP89.

src/main_nep/nep.cuh
```diff
@@ -96,6 +101,8 @@ public:
 
 private:
   ParaMB paramb;
+  ParaMB* d_paramb[16];
```
brucefan1983: do not use these.

src/main_nep/nep.cuh
```diff
@@ -61,11 +61,16 @@ public:
...
     int num_neurons1 = 0;           // number of neurons in the hidden layer
+    int num_neurons2 = 0;           // number of neurons in the output layer
+    int num_hidden_layers = 0;      // number of hidden layers
+    int one_ann_no_bias = 0;        // number of parameters in the ANN without bias
     int num_para = 0;               // number of parameters
     const float* w0[NUM_ELEMENTS]; // weight from the input layer to the hidden layer
-    const float* b0[NUM_ELEMENTS]; // bias for the hidden layer
-    const float* w1[NUM_ELEMENTS]; // weight from the hidden layer to the output layer
-    const float* b1;               // bias for the output layer
+    const float* b0[NUM_ELEMENTS]; // bias for the first hidden layer
```
brucefan1983: you can group `w0`, `b0`, `w1`, `b1`, `w2` as a single array `wb`

brucefan1983: In this way, you will save the constant memory for the kernels.

src/main_nep/parameters.cu
```diff
@@ -368,9 +373,18 @@ void Parameters::check_foundation_model()
     PRINT_INPUT_ERROR("Reading error for foundation model.");
   }
   if (num_neurons1 != get_int_from_token(tokens[1], __FILE__, __LINE__)) {
-    PRINT_INPUT_ERROR("neuron mismatches with foundation model.");
+    PRINT_INPUT_ERROR("the first layer neuron mismatches with foundation model.");
```
brucefan1983: restore this line.

src/main_nep/parameters.cu
```diff
@@ -368,9 +373,18 @@ void Parameters::check_foundation_model()
...
   }
 
+  if (num_neurons2 != get_int_from_token(tokens[2], __FILE__, __LINE__)) {
```
brucefan1983: remove these.

src/main_nep/nep.cuh
tang070205: That's a great idea. I'll try to incorporate it.

brucefan1983: you need to use the `wb` in the kernel with clever offsets, not mofiying the `device` functions

src/main_nep/parameters.cu
brucefan1983: you do not need to touch code related to the foundation model NEP89

tang070205: I just realized this is fine_tune. I've already restored it.

src/utilities/nep_utilities.cuh
```diff
@@ -160,7 +215,7 @@ static __device__ void apply_ann_one_layer(
   const float* w0,
```
brucefan1983, line 211: from here below, you don't need to touch, as the formal auguments do not need to be changed

src/main_nep/nep.cuh
tang070205: Revert to the original version and test passed

src/utilities/nep_utilities.cuh
tang070205: Got it. I'll revert it.

src/main_nep/nep.cuh
tang070205: Use “wb” as the default, and add parameters for the others

src/main_nep/nep_charge.cuh
```diff
@@ -44,7 +44,7 @@ public:
 
   struct ANN {
     int dim = 0;                    // dimension of the descriptor
-    int num_neurons1 = 0;           // number of neurons in the hidden layer
+    int num_neurons1 = 0;           // number of neurons in the first hidden layer
```
brucefan1983: this file should be restored.

tang070205: resolved

PR #1455, opened 2026-04-20
Optimize the VC-SGC MC trial loop in the NEP path

dingxu2016:
**Summary**

This PR adds two exact GPU-side optimizations to the `VC-SGC` MC trial loop in the `NEP` path. Tested on `RTX 4090` using binary alloy `VC-SGC` `NEP4` benchmarks, the combined patch improves cycle time by about `+16%` to `+20%` over the pre-optimization baseline, with no change to the `VC-SGC` acceptance rule.

**Modification**

1. Fuse before/after local `NEP` energy evaluation into one dual-state path and dispatch the angular contribution by `L_max`.
2. Build one global-shell neighbor list at the start of each `compute()` call and reuse it when constructing local inputs, replacing the previous per-trial all-atom scan.

Combined speedup vs the pre-optimization baseline:

| case | cycle speedup |
|---|---:|
| `N=4000, T=2000` | `+19.42%` |
| `N=4000, T=1500` | `+20.29%` |
| `N=4000, T=900` | `+16.32%` |
| `N=10000, T=2000, mc_trials=4000` | `+16.42%` |

**Others**

- No new external dependencies are introduced.
- Both changes are exact and keep the `VC-SGC` acceptance rule unchanged.
- Fixed-seed checks matched total `Delta E` and accept/reject behavior over the first 128 MC trials.
- For the input-construction change alone, paired `RTX 4090` runs improved cycle time by about `+6.81%` to `+9.33%` against the same dual-state + `L_max` path without global-shell input reuse.
- `nsys` shows the main benefit comes from local-input construction, while the dual-state `NEP` kernel stays broadly flat.
- The global-shell path is rebuilt once per `compute()` call to avoid persistent cross-call cache invalidation logic and keep the upstream patch reviewable.

### DIFF
--- a/src/mc/mc_ensemble_sgc.cu
+++ b/src/mc/mc_ensemble_sgc.cu
@@ -418,27 +418,15 @@ void MC_Ensemble_SGC::compute(
       z12_angular.data());
     GPU_CHECK_KERNEL
 
-    nep_energy.find_energy(
+    nep_energy.find_energy_dual(
       NN_ij_cpu,
       NN_radial.data(),
       NN_angular.data(),
       local_type_before.data(),
-      t2_radial_before.data(),
-      t2_angular_before.data(),
-      x12_radial.data(),
-      y12_radial.data(),
-      z12_radial.data(),
-      x12_angular.data(),
-      y12_angular.data(),
-      z12_angular.data(),
-      pe_before.data());
-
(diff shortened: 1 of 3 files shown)


brucefan1983:
I am not clear about the cycle speedup, what is it?

Could you provide a test input and compare the whole computation time?

PR #1463, opened 2026-04-23
Fix eam/alloy potential parsing

mushroomfire:
This PR fixes a bug in the eam_alloy potential file parsing. To ensure closer consistency with LAMMPS, the spline interpolation method has also been updated to match its implementation.

Validation has been performed by comparing energy, force, and virial against LAMMPS results. The differences are consistently within the 1e-4 level or better, which is likely limited by the use of float32 precision. Below are the test results for a CrCoNi eam/alloy system:
total E: lmp=-1332158.100253  gpumd=-1332158.097856
energy: max|diff|=1.751200e-05  rms=5.421647e-07
force : max|diff|=1.555827e-04  rms=5.918578e-06
virial: max|diff|=1.924350e-04  rms=5.169926e-06

Overall, this update improves compatibility and numerical consistency with LAMMPS for eam/alloy potentials.

### DIFF
--- a/src/force/eam_alloy.cu
+++ b/src/force/eam_alloy.cu
@@ -26,117 +26,61 @@
 #include <vector>
 #define BLOCK_SIZE_FORCE 64
 
-class CubicSpline
+// LAMMPS-style Hermite cubic spline using centered finite-difference derivatives.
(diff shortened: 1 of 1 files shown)


brucefan1983:
This is good for people caring about the consistency between codes, although it does not mean our previous version is wrong or not good.

PR #1464, opened 2026-04-23
Fix missing D_real mean correction in qNEP charge model

yongwangxx:
### Problem

The `zero_total_charge` kernel enforces charge neutrality by shifting predicted charges: `q_shift = q_pred - mean(q_pred)`. However, the corresponding chain-rule correction was missing in the force backpropagation pathway.

The electrostatic derivative `D_real = dU_ES/dq_shift` is consumed by `find_force_radial` and `find_force_angular` to compute forces via:

```
F = (Fp + charge_derivative * D_real) * d(descriptor)/dr
```

Since `q_shift` depends on all `q_pred` through the mean subtraction, the correct backpropagation requires:

```
D_real_corrected[i] = D_real[i] - mean(D_real)
```

Without this, the Jacobian of the charge neutralization projection (`δ_ij - 1/N`) is ignored, producing forces that are not the exact gradient of the energy.

### Impact

- **Non-conservative forces**: `F ≠ -∇U`, violating energy conservation
- **NVE energy drift**: systematic energy accumulation over time
- **Affects all qNEP systems**: especially slab, surface, cluster — geometry where D_real is not fully cancelled 

### Fix

Two new CUDA kernels, mirroring the structure of `zero_total_charge`:

| Kernel | File | Launch | Scope |
|--------|------|--------|-------|
| `zero_mean_D_real` | `src/force/nep_charge.cu` | `<<<1, 1024>>>` | Whole system (MD) |
| `zero_mean_D_real_train` | `src/main_nep/nep_charge.cu` | `<<<Nc, 1024>>>` | Per structure (training) |

Both use `double` shared-memory accumulators for numerical precision and are inserted after D_real is fully computed (Ewald/PPPM + real-space contributions) and before any force kernel reads it.

### Test Results

PbTe 2×2×2 supercell, `charge_mode=2`, 500,000 NVE steps (1 ns):

| Configuration | σ(E_total) / eV | vs Fixed |
|---------------|-----------------|----------|
| **Fixed (Ewald)** | 0.0002 | 1.0× |
| No charge (nep4) | 0.0002 | 1.0× |
| Original (Ewald, no fix) | 0.0009 | 3.8× |
| Original (PPPM, no fix) | 0.0015 | 6.6× |

### DIFF
--- a/src/force/nep_charge.cu
+++ b/src/force/nep_charge.cu
@@ -620,6 +620,41 @@ static __global__ void zero_total_charge(const int N, float* g_charge)
   }
 }
 
+
+// Chain rule correction: zero_total_charge shifted q by -mean(q),
+// so D_real must be shifted by -mean(D_real) for consistent forces.
+// Uses double accumulator for numerical precision.
+static __global__ void zero_mean_D_real(const int N, float* g_D_real)
+{
+  int tid = threadIdx.x;
+  int number_of_batches = (N - 1) / 1024 + 1;
+  __shared__ double s_sum[1024];
+  double sum = 0.0;
+  for (int batch = 0; batch < number_of_batches; ++batch) {
+    int n = tid + batch * 1024;
+    if (n < N) {
+      sum += (double)g_D_real[n];
+    }
+  }
+  s_sum[tid] = sum;
+  __syncthreads();
+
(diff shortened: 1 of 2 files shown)


brucefan1983:
@erhart1 This is the bug and the fix I mentioned in the email.

PR #1487, opened 2026-05-09
Update FIRE minimizer box constraint handling

ZhengZhong00:
**Summary**

Adds a `freeze_diagonal` parameter to the `minimize fire` command with box relaxation, allowing the user to keep one diagonal component of the box (xx, yy, or zz) fixed during minimization. This mimics LAMMPS' `fix box/relax` with a frozen coupled coefficient.

**Modification**
1. "src/minimize/minimizer_fire_box_change.cuh"
   - Merged two constructors into one with default parameters
   - Added `int freeze_diagonal` member (default -1 = no freeze)

2. "src/minimize/minimizer_fire_box_change.cu"
   - 5 freeze logic insertion points in `compute()`:
     1. Zero the virial component for the frozen diagonal 
     2. Exclude frozen diagonal from `max_stress_component` convergence check
     3. Zero the box velocity degree of freedom on GPU for frozen diagonal
     4. Restore frozen diagonal to original value after box update
   - Behavior unchanged when `freeze_diagonal < 0`

3. "src/minimize/minimize.cu"
   - Increased parameter count upper bound from 6 to 7
   - Added parsing for 7th parameter (valid: 0/1/2/3 for none/xx/yy/zz)
   - User-facing values 1/2/3 map internally to diagonal indices 0/4/8
   - Added print message for frozen diagonal case
 
**usage**
minimize fire <force_tolerance> <max_steps> <box_change> <hydrostatic_strain> <freeze_diagonal>

### DIFF
--- a/src/minimize/minimize.cu
+++ b/src/minimize/minimize.cu
@@ -67,8 +67,8 @@ void Minimize::parse_minimize(
   } else if (strcmp(param[1], "fire") == 0) {
     minimizer_type = 1;
 
-    if (!((num_param >= 4) && (num_param <= 6))) {
-      PRINT_INPUT_ERROR("minimize fire should have 2 to 4 parameters.");
+    if (!((num_param >= 4) && (num_param <= 7))) {
+      PRINT_INPUT_ERROR("minimize fire should have 2 to 5 parameters.");
     }
 
     if (!is_valid_real(param[2], &force_tolerance)) {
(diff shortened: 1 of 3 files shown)


brucefan1983:
The present extension does not seem to be very complete and mature.
I suggest you create a new sub-keyword such as `minimize fire_freeze`. In this way, your new code will not be coupled to the existing one and you have more degrees of freedom to modify it in the future.

PR #1538, opened 2026-06-02
Fix and document the HIP build on ROCm 7.x

jeffdaily:
GPUMD already ships a HIP build (`src/makefile.hip` plus the `gpu_macro.cuh` compat header), but it does not build on ROCm 7.x: rocPRIM/rocThrust now hard-require C++17 (the headers use `std::variant` / `std::is_same_v`) while `makefile.hip` inherited `-std=c++14` from the CUDA build, so the three Thrust-using translation units (`neighbor.cu`, `nep_multigpu.cu`, `dp.cu`) fail to compile.

This:
- bumps the HIP build to `-std=c++17`;
- makes the offload arch configurable (`HIP_ARCH ?= gfx90a`) so other AMD targets build with `make -f makefile.hip HIP_ARCH=<arch>` without editing the file;
- documents the HIP build in `doc/installation.rst`, which previously did not mention it (a prerequisites note plus a short "Building for AMD GPUs (ROCm/HIP)" section).

No `.cu`/`.cuh` or CUDA-build changes.

## Test plan

```bash
cd src && make -f makefile.hip clean && make -f makefile.hip -j HIP_ARCH=gfx90a
```

Validated on an AMD MI250X (gfx90a, ROCm 7.2.1) with three shipped MD examples: `carbon` (NEP4 ML potential, NVE) conserves total energy to 2.6e-7 relative drift and matches the committed CUDA reference PE to 5.8e-6; `graphene_dos` (Tersoff + phonon DOS via hipFFT, 20000 steps) conserves energy to 2.2e-6 and the DOS matches the CUDA reference; `graphene_kappa_emd` (Green-Kubo conductivity) runs clean. Run-to-run output is not bit-identical because the neighbor-list build uses atomicAdd ordering (non-deterministic on any GPU, CUDA included), but each run independently conserves energy and matches the reference, which is the correct gate for a chaotic MD trajectory.

This was prepared with the assistance of Claude (AI), reviewed before submission.

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -16,12 +16,28 @@ To compile (and run) :program:`GPUMD` one requires an Nvidia GPU card with compu
 On Linux systems, one also needs a C++ compiler supporting at least the C++11 standard.
 On Windows systems, one also needs the ``cl.exe`` compiler from Microsoft Visual Studio and a `64-bit version of make.exe <http://www.equation.com/servlet/equation.cmd?fa=make>`_.
 
+Alternatively, :program:`GPUMD` can be built for AMD GPUs through ROCm/HIP (see :ref:`Building for AMD GPUs <build_amd_hip>` below); this requires a ROCm installation.
+
 
 Compilation
 ===========
 
 In the ``src`` directory run ``make``, which generates two executables, ``nep`` and ``gpumd``.
 Please check the comments in the beginning of the makefile for some compiling options.
+
+
+.. _build_amd_hip:
+
+Building for AMD GPUs (ROCm/HIP)
+================================
+
+:program:`GPUMD` also runs on AMD GPUs through ROCm/HIP. In the ``src`` directory, build with the HIP makefile instead of the default one::
+
+  make -f makefile.hip
+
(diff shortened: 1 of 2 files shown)


tang070205:
This is not a problem for ROCM6.

jeffdaily:
> This is not a problem for ROCM6.

It is for 7.  Would it be acceptable to make this backward compat with rocm 6?

jeffdaily:
C++14 is fine on ROCm 6. The issue is that pinning -std forces it down: on ROCm 7.x rocThrust/rocPRIM require C++17 (rocPRIM emits `#error "rocPRIM requires at least C++17"`), and hipcc already defaults to C++17. Rather than bump the pin, I dropped -std from makefile.hip and let the compiler default apply, so it builds on both ROCm 6 and 7 with no version-specific flag; the CUDA makefile's -std=c++14 is untouched.

brucefan1983:
The changes look safe and good to me. When you make sure it still works for our AMD card we can merge @tang070205

jeffdaily:
Which card?

tang070205:
> Which card?  哪张卡？

 @jeffdaily  7900XT

jeffdaily:
[7900XT is gfx1100](https://rocm.docs.amd.com/projects/install-on-windows/en/latest/reference/system-requirements.html), which is what I tested.

PR #1567, opened 2026-06-21
Fix: correct PPPM per-atom virial handling for k=0 and virial-dependent outputs

cmdhwz:
**Summary**

This PR fixes two PPPM correctness issues in GPUMD while keeping all existing input syntax unchanged. It ensures that PPPM produces consistent per-atom virials for outputs that explicitly depend on them, such as compute ... virial, compute ... jp, and dump_xyz ... virial.

**Modification**

1. The k = 0 entry in the PPPM per-atom virial mesh is now explicitly reset to zero. This prevents stale data from previous steps from being reused when mesh_virial is updated in place and later overwritten by inverse FFT output.

2. The logic that decides whether PPPM must compute true per-atom virials has been extended. Previously, some commands that consume per-atom virials could still fall back to the simplified path that distributes the total virial evenly across atoms. This PR makes compute ... virial, compute ... jp, and dump_xyz ... virial correctly trigger the full per-atom PPPM virial path.

**Others**

Thermal conductivity calculations using qNEP models trained for BaTiO3 and AlAs have been tested. In PPPM mode with a mesh spacing of 1 Angstrom, BaTiO3 gives results that are in very good agreement with Ewald mode. For AlAs, the calculated thermal conductivity is also consistent with the BTE thermal conductivity obtained with Born effective charge corrections.

### DIFF
--- a/src/utilities/read_file.cu
+++ b/src/utilities/read_file.cu
@@ -126,6 +126,28 @@ bool check_need_peratom_virial()
         need_peratom_virial = true;
         break;
       }
+      if (tokens[0] == "compute") {
+        for (const auto& token : tokens) {
+          if (token == "virial" || token == "jp") {
+            need_peratom_virial = true;
+            break;
+          }
+        }
+        if (need_peratom_virial) {
+          break;
+        }
+      }
+      if (tokens[0] == "dump_xyz") {
+        for (const auto& token : tokens) {
+          if (token == "virial") {
+            need_peratom_virial = true;
+            break;
+          }
+        }
+        if (need_peratom_virial) {
+          break;
+        }
+      }
     }
   }
   input_run.close();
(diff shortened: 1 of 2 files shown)


src/utilities/read_file.cu
```diff
@@ -126,6 +126,28 @@ bool check_need_peratom_virial()
...
         break;
       }
+      if (tokens[0] == "compute") {
+        for (const auto& token : tokens) {
+          if (token == "virial" || token == "jp") {
+            need_peratom_virial = true;
+            break;
+          }
+        }
+        if (need_peratom_virial) {
```
brucefan1983, line 136: this seems to be redundant

src/utilities/read_file.cu
```diff
@@ -126,6 +126,28 @@ bool check_need_peratom_virial()
...
+        }
+      }
+      if (tokens[0] == "dump_xyz") {
+        for (const auto& token : tokens) {
+          if (token == "virial") {
+            need_peratom_virial = true;
+            break;
+          }
+        }
+        if (need_peratom_virial) {
```
brucefan1983, line 147: this seems to be redundant

brucefan1983:
Nice fix! Thermal transport is alive.

PR #1579, opened 2026-06-30
Add more information to several output files

erhart1:
This PR adds headers to several output files analogously to #1552, specifically including
* `thermo.out` (`dump_thermo`)
* `msd.out` (`compute_msd`)
* `viscosity.out` (`compute_viscosity`)
* `sdc.out` (`compute_sdc`)

This makes parsing these files simpler and much less error prone since information such as the number of atoms, the cell metric, and the time step or dump frequency are included in the file and do not have to provided externally.

Each header records the originating command and its call signature, `format_version`, `num_atoms`, cell metric (where not already in every data row), the time spacing between output rows, and column names.
The headers are compatible with numpy/pandas comment-skipping parsers.

### DIFF
--- a/src/makefile
+++ b/src/makefile
@@ -15,7 +15,7 @@
 # some flags
 ###########################################################
 CC = nvcc
-CUDA_ARCH=-arch=sm_60
+CUDA_ARCH=-arch=sm_80
 WARN_FLAGS = -Wall -Wextra
 WARN_FLAGS += -Wno-sign-compare -Wno-unused-parameter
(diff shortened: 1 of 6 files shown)


brucefan1983:
Thanks, I will let some users to know this first, and see if there is any question.

erhart1:
> Thanks, I will let some users to know this first, and see if there is any question.

Sounds good.
I added `shc.out` in a separate PR (#1581) since that file format is a bit special.

Issue #1582, opened 2026-07-01
`compute_viscosity` produces `inf` running integrals in NVE because Green-Kubo prefactor uses zero thermostat temperature


kylinguo2016:
### Summary

When `compute_viscosity` is used in an NVE-only input (e.g., a continuation run from `restart.xyz`), columns 11-19 of `viscosity.out` can be `inf`, while stress autocorrelation columns 2-10 remain valid.

This was found in GPUMD 5.3 and is still present on current `master`. It appears to be a boundary case: GPUMD expects users to define temperature via a thermostatted ensemble (NVT/NPT) in the same input before NVE production. Using NVE-only input—including NVE for short re-equilibration—is uncommon and was likely not handled explicitly.

### Expected developer workflow

```
ensemble    nvt_lan 2000 2000 100   # temperature well defined
run         ...

ensemble    nve
compute_viscosity ...
run         ...                     # production; temperature2 may carry over
```

### Boundary workflow (this issue)

```
# Continuation from restart.xyz; no thermostat in this input file
ensemble    nve                     # optional short NVE re-equilibration (unusual!!!)
run         ...

ensemble    nve
compute_viscosity 20 4000
run         ...                     # temperature2 never set → prefactor breaks
```

NVE has no thermostat and does not define a target temperature. If no prior thermostatted block exists in the current `run.in`, `integrate.temperature2` remains zero and the Green-Kubo prefactor in `Viscosity::postprocess()` divides by zero:

```
double factor = dt * 0.5 / (K_B * temperature * box.get_volume());
```

Note: the system may still be physically equilibrated (e.g., from a previous NVT job), but GPUMD does not infer that temperature for the viscosity integration.

### Steps to Reproduce

1. Equilibrate with NVT in one job (or omit thermostat entirely in current input).
2. Start a new GPUMD run with NVE-only `run.in` and `compute_viscosity`.
3. Inspect columns 11-19 of `viscosity.out`.

### Expected vs actual

|                           | Expected | Actual   |
| :------------------------ | :------- | :------- |
| Columns 2-10 (stress ACF) | Finite   | Finite ✓ |
| Columns 11-19 (running η) | Finite   | `inf` ✗  |

### Impact

Affects NVE-only or continuation workflows without a thermostatted block in the same input. The standard single-file NVT → NVE workflow may appear to work if `temperature2` persists from the NVT block.

### Suggested fix

[...]

PR #1583, opened 2026-07-01
fix compute_viscosity

kylinguo2016:
**Summary**

Improve #1582 `compute_viscosity` handling when NVE production is run without a thermostatted ensemble in the same `run.in`.

This edge case was found in GPUMD 5.3 during Green-Kubo viscosity work and remains on current `master`. GPUMD’s expected workflow is NVT/NPT equilibration (temperature defined) followed by NVE production. In our case, a continuation run used NVE-only input—including a short NVE re-equilibration stage—without any NVT/NPT block in that file. Although the structure was equilibrated in a prior job, `Viscosity::postprocess()` still uses `integrate.temperature2`, which is zero for NVE-only inputs, causing division by zero and `inf` in `viscosity.out` columns 11-19.

This PR makes the Green-Kubo prefactor use the time-averaged kinetic temperature sampled during production, so continuation and NVE-only workflows behave correctly without changing the standard NVT → NVE path.

**Modification**

- `src/measure/viscosity.cuh`: Add accumulators for kinetic temperature during stress sampling.
- `src/measure/viscosity.cu`:
  - Accumulate kinetic temperature from `thermo[0]` in `process()`.
  - Use the averaged kinetic temperature in `postprocess()` for the Green-Kubo prefactor.
  - For PIMD (`integrate.type >= 31`), keep using the passed thermostat temperature.
  - Emit a diagnostic when NVE production relies on kinetic rather than thermostat target temperature.
  - Error if no valid temperature is available.

**Others**

- Scope: Primarily affects NVE-only / continuation workflows. Single-file NVT → NVE runs are largely unchanged.
- Validation: NVE-only `compute_viscosity 20 4000` run — unfixed code gives `inf` in columns 11-19; with this fix, all columns are finite and match re-integration from columns 2-10.
- Design note: NVE is a poor choice for equilibration; this fix does not endorse that workflow, but prevents silent failure when users run `compute_viscosity` without a thermostat-defined temperature in the current input.
- Output format: Unchanged; only numerical values of columns 11-19 are corrected in affected cases.

**Discussion**

We originally hit this while running long Green–Kubo viscosity calculations on a molten system with GPUMD 5.3. Our main equilibration path follows the usual GPUMD workflow (NVT/NPT before production), but for a continuation job we used an NVE-only `run.in`—including a short NVE re-equilibration stage—without a thermostatted block in that same input file. In that setup, `viscosity.out` columns 2–10 looked fine, while columns 11–19 overflowed to `inf`, which traced back to the Green–Kubo prefactor relying on a thermostat-defined temperature that NVE does not provide in the current run context. We understand this is an uncommon boundary case rather than the primary intended path, and we are not suggesting a change to the recommended NVT/NPT → NVE workflow; we simply thought it would be helpful to fail gracefully here by using the sampled kinetic temperature during production. We would be happy to adjust the implementation or wording if you prefer a different behavior for continuation runs. Thank you for maintaining GPUMD and for considering this small robustness improvement.

Files used to test.
[share.tar.gz](https://github.com/user-attachments/files/29538374/share.tar.gz)

### DIFF
--- a/src/measure/viscosity.cu
+++ b/src/measure/viscosity.cu
@@ -18,8 +18,10 @@ Calculate the stress autocorrelation function and viscosity.
 ------------------------------------------------------------------------------*/
 
 #include "utilities/common.cuh"
+#include "utilities/error.cuh"
 #include "utilities/gpu_macro.cuh"
 #include "utilities/read_file.cuh"
+#include "integrate/integrate.cuh"
 #include "viscosity.cuh"
 #include <vector>
 #include <cstring>
@@ -38,6 +40,8 @@ void Viscosity::preprocess(
   if (compute) {
     int number_of_frames = number_of_steps / sample_interval;
     stress_all.resize(NUM_OF_COMPONENTS * number_of_frames);
+    accumulated_kinetic_temperature_ = 0.0;
+    kinetic_temperature_sample_count_ = 0;
   }
 }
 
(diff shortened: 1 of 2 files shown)


brucefan1983:
As we discussed in QQ, this is due to the absence of an NVT/NPT before NVE. It is a feature, not a bug.

PR #1586, opened 2026-07-02
Add test suite based on pytest and calorine

erhart1:
### Summary

Adds a new, independent pytest test suite under `tests_pytest/`, built on calorine's `GPUNEP`
ASE calculator driving the real `gpumd` binary. Complements the existing manual `tests/`
suite without touching it. It is a parallel test infrastructure. There are no changes to GPUMD's
source code.

### What is covered

- **Physics correctness**: translation/rotation/permutation/PBC invariance, analytic-vs-
  finite-difference force consistency, NVE energy/momentum conservation, GPUMD's own
  PPPM-vs-Ewald reciprocal-space agreement (relative check), and rotational self-consistency for
  TNEP tensor predictions (dipole rotates as a vector, polarizability transforms as a rank-2
  tensor).
- **Golden-file regression**: `test_regression.py`, tolerance-based comparisons frozen in
  `fixtures/golden/*.npz` and regenerated only via `--update-golden` — covers a BaZrO3
  system (energy/forces) and Born effective charges for both of GPUMD's reciprocal-space methods
  (Ewald, PPPM) independently, so a regression in either implementation is caught, not just
  their mutual disagreement.
- **Command IO smoke tests** ("Tier 1"): `test_io_basic_setup.py`, `test_io_dump_commands.py`,
  `test_io_compute_commands.py`, `test_io_tnep_commands.py` — confirm ~35 `run.in` keywords
  execute cleanly and produce output of the expected shape, across the `structure` × `model_type`
  fixture matrix (`bulk_C`, `bulk_perovskite`, `bulk_water` × `nep`, `qnep_mode1`, `qnep_mode2`),
  with TNEP (`dump_dipole`/`dump_polarizability`) as fixed (structure, PES, TNEP-observer)
  triples outside that matrix, including their own golden-file regression.
- **I/O parsing**: `test_parsing.py` — round-trip and hypothesis-based malformed-input tests for
  model files and fixed-width output files, no GPU required.

Markers: `fast` (default), `slow` (MD-conservation-scale, excluded by `-m "not slow"`).

### Approach

Numeric correctness is judged against this repo's own previously-generated output — golden
files, or GPUMD evaluated under two of its own settings against each other (e.g. PPPM vs. Ewald).

In addition there is a standalone developer utility at `scripts/compare_gpunep_cpunep.py` which
can be used to validate the GPU implementation in `gpumd` against the CPU implementation that
is part of calorine.

qNEP models default to `kspace ewald` (`conftest.py`'s `make_gpunep`) in comparisons
that would otherwise conflate the reciprocal-space method with whatever else is being checked
(e.g. rotation invariance, where PPPM's fixed-orientation FFT mesh would introduce an unrelated
artifact); GPUMD's own PPPM-vs-Ewald behavior is validated on its own terms elsewhere
(`test_kspace_consistency.py`'s relative check, `test_regression.py`'s independent golden files
for each method).

### Fixtures

Publishd NEP/qNEP/TNEP models (water, diamond C, BaTiO3, BaZrO3) at small
system sizes.

### Some potential bugs that were identified while building the suite

- `change_box <delta>` hangs indefinitely on boxes with zero on-diagonal lattice components.
- `compute`/`compute_adf` silently write an empty output file (no error) if the sample/output
  interval exceeds the run length.

### Not done yet

- Not integrated in CI, Tier 2/3 command IO smoke tests, multi-GPU decomposition consistency.
- Golden-file regression currently covers one full-size system (BaZrO3) plus BEC; plain-NEP/qNEP
  energy-forces golden regression on the smaller swept structures is intentionally *not*
  duplicated there (invariance/finite-difference self-consistency is this suite's validation
  mechanism for those), but could be extended with the same both-kspace-methods pattern later if
  wanted.

### How to run

Provided pytest and calorine are available the suite can be run via `pytest tests_pytest`.
On my workstation (RTX3080) running the suite takes about 4.5 minutes.
332 tests pass, 97 are skipped, none fail.
The ones that are skipped are due to combinations that are not accessible, e.g., there is no qNEP model for C.
This is part of the design, which creates different combinations of models and tests.


erhart1:
@brucefan1983 this is an attempt at setting up the test suite of the `gpumd` interface based on `pytest` and `calorine` that I outlined before. Finally got around to setting it up.

tests_pytest/conftest.py
```diff
@@ -0,0 +1,277 @@
...
+_STRUCTURE_BUILDERS = {
+    'bulk_C': make_bulk_C,
+    'bulk_perovskite': make_bulk_perovskite,
+    'bulk_water': make_bulk_water,
+}
+
+# (structure_name, model_type) -> file in fixtures/models/. Missing combinations (currently
+# bulk_C + either qnep mode, since no qNEP model was supplied for carbon) are skipped
+# explicitly in model_path below rather than silently omitted from the params list.
+_MODEL_FILES = {
```
elindgren, line 186: Why are the TNEP models skipped here? Is it due to all of these models being able to predict energies/forces?

tests_pytest/test_io_compute_commands.py
```diff
@@ -0,0 +1,168 @@
...
+    expected_output_files=['elastic.out'], parse_check=_check_elastic_format)
+
+
+@pytest.mark.parametrize(
+    'case', [*COMPUTE_COMMAND_CASES, COMPUTE_ELASTIC_CASE], ids=lambda c: c.name)
+def test_command_io(tmp_path, structure, structure_name, model_path, model_type, gpumd_command,
+                     case):
+    if case.name == 'compute_elastic' and structure_name == 'bulk_water':
+        pytest.skip(
+            'compute_elastic derives elastic constants from the stress response to small '
```
elindgren: This, and the skip in `test_compute_phonon` are very verbose. Fine, but maybe a bit excessive

.gitignore
```diff
@@ -26,5 +26,7 @@ public.*
 
 __pycache__/
 */__pycache__/
+.pytest_cache/
+.hypothesis/
```
elindgren, line 30: what is a .hypothesis file?

elindgren (commented):
Very nice! We could potentially get this into the actual CI, perhaps only running the fast tests on push and then the slow tests on merge to master.
We could maybe use larger github runners for access to GPUs: https://docs.github.com/en/actions/concepts/runners/larger-runners

brucefan1983:
Thanks Paul @erhart1 
I'll ask one of my students to play with it and see if any questions come up. I'm totally disabled with Python, so I need help :-)

BBBuZHIDAO:
Very nice work! I think this will be very helpful for GPUMD development, especially for making future testing more standardized and convenient. I tested it locally on RTX 3060 Ti , and I got the same result as reported:

```bash
pytest tests_pytest -q
# 332 passed, 97 skipped in about 4.7 min
```

I only have two small suggestions from my local setup experience:

1. For the executable path, the tests seem to expect `gpumd` to be located at the repository root, while the normal build generates it under `src/gpumd`. For example, the current setting is:

```python
GPUMD_EXECUTABLE = REPO_ROOT / "gpumd"
```

Would it be possible to first look for `gpumd` under the repository root, and if it does not exist, fall back to `REPO_ROOT / "src" / "gpumd"`? This would avoid the need to manually create a symlink after building.

2. It would also be helpful to add a short README for this test suite, including the basic setup and run instructions, such as the required Python packages/environment, how to build `gpumd`, how the executable path is detected, and the recommended commands for running the fast/full tests.

Overall, this is a very useful and well-structured testing framework, and it runs successfully on my side.

tests_pytest/conftest.py
erhart1: The tests are set up as a matrix, where the available models are run against different tests.
TNEP models do not provide energy and forces and cannot be used for such tests.

.gitignore
erhart1: See [here](https://hypothesis.readthedocs.io/en/latest/)

erhart1:
Happy to hear about the positive feedback!
The default location for the gpumd binary is now `src/gpumd` (d48d23b8d2f27c) and I added a `README.md` file to the `tests_pytest` directory.

Note that this test suite is not complete yet (e.g., quite a few ensemble keywords are missing and `nep` is untested) but this should be handled in additional PRs.

tests_pytest/test_io_compute_commands.py
erhart1: Addressed in 795d1abd

PR #1605, opened 2026-07-10
Allow user to override the q-scaler reset

erhart1:
This PR introduces the `import_q_scaler` input parameter for `nep`.If set to `1` the `q_scaler` will taken from the `nep.txt` file available in the run directory. This allows a user to override the reinitialization of the `q_scaler` on restart. This behavior is necessary for truthful fine-tuning based on externally provided restart files. The `fine_tune` command does this already internally.


erhart1:
FYI @elindgren

brucefan1983:
It seems this PR assumed the local (small) model to be fine-tuned is from NEP89, but I will merge it and let extension to be done in the future. @erhart1

PR #1608, opened 2026-07-10
Add output interval control for nep

erhart1:
This PR introduces the `output_interval` parameter, which allows one to set the dump frequency to `loss.out` directly. The default is 100 as before, so not setting the keyword recovers the known behavior. The keyword is meant as an expert option that allow users to inspect the early stages of a training run, in particular during fine tuning.


erhart1:
ping @elindgren

PR #1610, opened 2026-07-10
Removed unused q_scaler keyword from nep.in

erhart1:
While working on #1605 I noticed that the `nep.in` parser parses the `q_scaler` keyword but that this command is not used anywhere in the code. It was presumably introduced for testing purpose to allow a user to manually set the `q_scaler` but the actual functionality was later removed while the keyword remained. This PR removes this unused keyword.


brucefan1983:
After merging the previous two PRs, this PR has conflicts with master. Could you resolve it? @erhart1

PR #1627, opened 2026-07-21
add lincs

tang070205:
This keyword is used to constrain bond lengths using the LINCS (linear constraint solver) algorithm [Hess1997].
With the fastest (bond-stretching) degrees of freedom frozen, a larger time step can be used in molecular dynamics simulations, for example, 1 fs or more for systems containing hydrogen atoms.

For automatic detection, a pair of atoms `i` and `j` is considered bonded when
`   r_{ij} < f \left( r_i^{\rm cov} + r_j^{\rm cov} \right)`
where `f` is the ``factor`` parameter described below.

This keyword is used as follows:
  lincs <mode> [order <order>] [iter <iter>] [factor <factor>]

The `mode` can be:
  - `H` (default): only constrain bonds involving hydrogen atoms;
  - `all`: constrain all bonds detected from the initial structure;
  - `file <filename>`: read the constraints from the file `filename` (see below for the format);
  - `off`: disable bond constraints.

The parameters that follow: 
  - `order` is the order of the Neumann series expansion in LINCS, with a default value of 4.
  - `iter` is the number of re-linearization iterations per step, with a default value of 1.
  - `factor` is the factor `f` in the bond detection criterion above, with a default value of 1.25. It is not used with the `file` mode.

The constraint file uses atom indices starting from 1.
Each non-comment line contains two atom indices and an optional target length in units of Å:
```
  # atom_i atom_j [target_length_in_A]
  1 43
  2 43 0.96
```

Examples
--------
Constrain the bonds involving hydrogen atoms with default parameters:
`lincs H`

Constrain all the detected bonds, with an expansion order of 6:
`lincs all order 6`

Read the constraints from the file `constraints.in`
`lincs file constraints.in order 6 iter 2`


brucefan1983:
This PR is not matured and we also might not need this functionality.

PR #1640, opened 2026-07-29
Fix handling of netcdf/HDF5 linking

erhart1:
## Summary

When linking against a netcdf library that has been compiled without HDF5 and compression support the compilation fails. This occurs after the addition of support for HDF5 and compression in #1631. While the documentation states that the compression is optional the code calls functions (e.g., `nc_def_var_chunking`) that are only available in netcdf if it has been compiled with HDF5 support.

This PR adds a preprocessor if-statement that only includes these calls at link/compile time if HDF5 support is actually available. It also adds a short clarifying statement to the documentation of the `dump_netcdf` command.

## Additional note
On  my system (ubuntu) several additional steps were required to compile netcdf with HDF5 and compression support.
I did not add any of the information below to the installation page because I am unsure about how generalizable it is.
But one should consider where this information could or should land.
* I had to install HDF5 with development headers as well as `m4` (apt install libhdf5-dev m4).
* The HDF5 headers/libraries ended up under a non-standard path (`/usr/include/hdf5/serial` and `/usr/lib/<arch>/hdf5/serial`), which meant that `configure` had to be pointed at these path explicitly
  ```
   CPPFLAGS=-I/usr/include/hdf5/serial LDFLAGS=-L/usr/lib/x86_64-linux-gnu/hdf5/serial
   ./configure --prefix=<path> --enable-netcdf-4 --disable-dap
  ```
* One should note that `configure` silently falls back to `--disable-netcdf-4` if it cannot find a working HDF5 installation. To check that HDF5 is actually available one can run `bin/nc-config --has-hdf5` in the netcdf base directory.
* Furthermore I had to include the HDF5 paths and libraries in the `makefile`
  ```
  INC = -I<path>/include -I./ -I<hdf5_path>/include
  LDFLAGS = -L<path>/lib -Xlinker=-rpath -Xlinker=<path>/lib -L<hdf5_path>/lib
  LIBS = -lcublas -lcusolver -lcufft -lnetcdf -lhdf5_hl -lhdf5 -lz
  ```

### DIFF
--- a/doc/gpumd/input_parameters/dump_netcdf.rst
+++ b/doc/gpumd/input_parameters/dump_netcdf.rst
@@ -45,6 +45,13 @@ Currently, the following optional arguments are accepted:
     Level 0 applies no compression. For large, frequently written trajectories, use ``none`` for
     maximum write speed, ``deflate 1`` for a practical balance, or ``deflate 9`` when minimizing
     file size is more important than write speed.
+    GPUMD itself does not require NetCDF4/HDF5 support unless ``deflate`` is used.
+    If the linked NetCDF-C library was built without NetCDF4/HDF5 support, requesting
+    ``deflate`` causes the underlying ``nc_create`` call to fail when GPUMD asks for the
+    NetCDF4 file format.
+    GPUMD already checks for this and reports a clear error instead of silently writing an
+    uncompressed file.
+    See :ref:`NetCDF setup <netcdf_setup>` for how to build NetCDF-C with NetCDF4/HDF5 support.
 
 Requirements and specifications
 -------------------------------
--- a/src/measure/dump_netcdf.cu
+++ b/src/measure/dump_netcdf.cu
@@ -35,6 +35,7 @@ https://ambermd.org/netcdf/nctraj.pdf
 #include "model/box.cuh"
 #include "model/group.cuh"
 #include "netcdf.h"
+#include "netcdf_meta.h"
 #include "parse_utilities.cuh"
 #include "utilities/common.cuh"
 #include "utilities/error.cuh"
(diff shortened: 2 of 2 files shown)


Tingliangstu:
Thank you, Paul, for identifying and fixing this compatibility issue. The change makes the optional NetCDF-4/HDF5 dependency behave as originally intended in #1631: a classic NetCDF build can still be used with `compression none`, while deflate compression is available when HDF5 support is enabled.

Would it be slightly more robust to write the condition as

#if defined(NC_HAS_HDF5) && NC_HAS_HDF5

instead of

#if NC_HAS_HDF5

This would also avoid warnings with `-Wundef` if the macro is unavailable for some reason. Since `netcdf_meta.h` should define the macro, I consider this a small robustness improvement.

Regarding the installation notes, I agree that the HDF5 paths are distribution- and installation-dependent. On my Ubuntu system, NetCDF-C 4.10.1 was built as a shared library with HDF5 support, and the following GPUMD configuration works without explicitly linking HDF5 and zlib:

INC = -I/xxx/netcdf-c-4.10.1/build/include -I./
LDFLAGS = -L/xxx/netcdf-c-4.10.1/build/lib \
          -Xlinker -rpath \
          -Xlinker /xxx/netcdf-c-4.10.1/build/lib
LIBS = -lcublas -lcusolver -lcufft -lnetcdf

Here, `-L` locates `libnetcdf.so`, and the shared NetCDF library carries its HDF5 and zlib dependencies. Other systems or static builds may require additional libraries.

Because the exact HDF5 include and library paths vary considerably between systems, I suggest keeping the general installation documentation as it is instead of adding Ubuntu-specific paths. Users who cannot build NetCDF-4/HDF5 support can still use a classic NetCDF build with `compression none`. Moreover, with the assistance of LLM, users can usually diagnose and resolve system-specific installation issues more easily than before. Therefore, I suggest keeping the current installation documentation unchanged for now.

Thanks again for fixing this bug.

erhart1:
> Would it be slightly more robust to write the condition as
> 
> #if defined(NC_HAS_HDF5) && NC_HAS_HDF5
> 
> instead of
> 
> #if NC_HAS_HDF5
> 
> This would also avoid warnings with -Wundef if the macro is unavailable for some reason. Since netcdf_meta.h should define the macro, I consider this a small robustness improvement.

That's a good suggestion. Added in 7ad53e526e528b9


> Regarding the installation notes, I agree that the HDF5 paths are distribution- and installation-dependent. 

Great. Then we are on the same page. Anything else to consider?

PR #1643, opened 2026-07-31
Fix test_nep_model_consistency after the nep.restart interval change

erhart1:
Since `d945828c`, `SNES::compute` writes `nep.restart` every output_interval generations instead of every 100. The `train_once()` helper in `tests_pytest/test_nep_model_consistency.py` runs `nep` with `output_interval 1`, so its
short training run now leaves a `nep.restart` in the directory. That made the second run in `test_stale_nep_txt_without_restart_only_warns` a resume rather than a retrain in place, so it took the fatal "remove nep.restart" branch of `Parameters::check_existing_model` instead of the non-fatal "nep.txt will be overwritten" warning the test asserts on.

`train_once()` now deletes the `nep.restart` it produces. Every test that wants one already writes its own with `write_nep_restart()`, so no test loses coverage. The module docstring and the comment above `COMMON_KEYWORDS`, which both still described the old fixed interval of 100 generations, are corrected.

### DIFF
--- a/tests_pytest/test_nep_model_consistency.py
+++ b/tests_pytest/test_nep_model_consistency.py
@@ -2,11 +2,11 @@
 directory.
 
 A normal training run writes nep.txt, so a second invocation of `nep` in the same directory always
-finds one, and a resume additionally finds the nep.restart written every 100 generations. If the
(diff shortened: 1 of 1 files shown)


erhart1:
Sorry I had missed this in the previous PR.

PR #1646, opened 2026-08-06
Rework the dump_xyz syntax and remove four outdated dump keywords

erhart1:
## Summary

`dump_position`, `dump_velocity`, `dump_force`, and `dump_exyz` each wrote a fixed set of quantities to a hard-coded file name, and each carried its own parsing, GPU copy kernel, and file handling. All four are covered by dump_xyz, so they are removed and now report the `dump_xyz` invocation that replaces them.

`dump_xyz` itself is reworked while its documentation was being touched anyway:

- The grouping method and group ID move from the first two positional parameters to an optional `group <m> <id>`, matching dos, sdc, shc, and msd. When the old format is used an error is printed that redirects the user to the new format.
- The quantity keyword `group` becomes `group_labels`, since `group` is now the selection option.
- A `precision single|double` option is added, as this was one of the features of `dump_position` that was not yet available via `dump_xyz`. The same behavior is already supported by `dump_netcdf`. The previous output was a fixed %.8f. For consistency with the other precision keywords the new default is single precision, meaning nine significant digits.
- Unrecognized arguments trigger an error message and the program execution to terminate.

## Breaking changes

Input files using the four removed keywords, or the old `dump_xyz` argument order, need updating. Every case reports which command to use instead and then stops.

## Testing
Extended the `dump_xyz` coverage and dropped the legacy reference files `movie1.xyz`, `force1.out`, `dump1.xyz`, and `velocity1.out` that were used by the `drop_*` commands that were removed.

## Notes
* The error messages concerning the old keywords should be retired with the next major release.
* I have not been able to check the MDI version since it is available in my environment.
* I think it would be nice if the call signature used here could be mirrored for the `dump_netcdf` command.
* The `dump_observer` and `dump_beads` commands duplicate some functionality of `dump_xyz`. They could be refactored to reuse `dump_xyz` functionality. Taking this even further the two commands could be set up to allow users to choose between netcdf and xyz.


erhart1:
@brucefan1983 here is a PR that removes the `dump_*` commands as discussed elsewhere. It would be great if some more people could give a try. To prevent premature merging I therefore marked it as "draft".

Note that the changes look very extensive. That is largely because a lot of commands are being removed (over 34,000 lines!).

src/main_gpumd/deposition.cu
```diff
@@ -262,8 +262,8 @@ void Deposition::analyze_run(const std::string& filename)
 
       if (int(n) == run_line) {
         lines.emplace_back(
-          "dump_xyz -1 0 " + std::to_string(interval) + " deposited_" +
-          std::to_string(dump_index) + ".xyz velocity" + (has_group ? " group" : ""));
+          "dump_xyz " + std::to_string(interval) + " deposited_" + std::to_string(dump_index) +
+          ".xyz velocity" + (has_group ? " group_labels" : ""));
```
brucefan1983, line 266: `group_labels` should be changed back to `group`?

erhart1: `group` is now a keyword itself that allows a user select which groups to include in the output (previously given by mandatory arguments, which made the line more opaque). Therefore the property had to be renamed.

```
dump_xyz 20 dump.xyz                  # 1) write full system to dump.xyz every 20 steps
dump_xyz 20 dump.xyz group_labels     # 2) as 1 but include group label
dump_xyz 20 dump.xyz group 0 1        # 3) as 1 but only include atoms in group 1 of grouping method 0
dump_xyz 20 dump.xyz group 0 1 force  # 4) as 3 and include forces in output
```

erhart1: If one runs `dump_xyz 20 dump.xyz group` one gets an error message that specifies how to fix the command:
```
Input Error:
    File:       measure/dump_xyz.cu
    Line:       123
    Error text: Option 'group' should be followed by a grouping method and a group ID. The quantity that writes group labels as a column is now called 'group_labels'.
```

erhart1: If this PR gets merged I suggest to create an issue (task) to be handled for the next major release (5.X), namely that the various error messages (also for the now removed commands) should be removed.

brucefan1983: I see, the `dump_xyz` keyword has been changed quite a lot.

PR #1648, opened 2026-08-08
Revise and extend the `dump_netcdf` command

erhart1:
Brings `dump_netcdf` in line with `dump_xyz`, in two parts.

## Arguments

The output interval and the file name are now the only positional arguments:

```
dump_netcdf <interval> <filename> [group <m> <id>] [precision single|double]
            [compression none|deflate <level>] {<quantity> ...}
```

- the grouping method and group ID move into a `group` option parsed by `parse_group`, replacing inline bounds checks
- `has_velocity` becomes the optional quantity `velocity`
- unrecognized arguments are rejected instead of ignored
- an input file written for the old positional form is detected and reported with the replacement spelled out
  
Options and quantities may be given in any order.

## Quantities

All nine per-atom quantities `dump_xyz` writes are now available, so an input line that works for one works for the other:

| quantity | variable | dimensions | GPUMD unit |
| --- | --- | --- | --- |
| `velocity` | `velocities` | `(frame, atom, spatial)` | Å/ps |
| `force` | `forces` | `(frame, atom, spatial)` | eV/Å | 
| `unwrapped_position` | `unwrapped_coordinates` | `(frame, atom, spatial)` | Å |
| `potential` | `potential_energy` | `(frame, atom)` | eV | 
| `charge` | `charge` | `(frame, atom)` | e | 
| `bec` | `bec` | `(frame, atom, spatial, spatial)` | e |
| `virial` | `virial` | `(frame, atom, spatial, spatial)` | eV |
| `mass` | `mass` | `(atom)` | amu |
| `group_labels` | `group_labels` | `(atom, grouping_method)` | -- |

Mass and group labels cannot change during a run and are written once, without a frame dimension.
The requested set is recorded as the `gpumd_quantities` attribute, which replaces `gpumd_has_velocity` and is what appending compares.

Quantities that carry direction follow the rotation into the restricted AMBER cell that positions and velocities already went through: vectors as `R v`, rank-2 tensors as `R T Rᵀ`.
Without this they would describe a different frame from the coordinates beside them.

Forces are written in the units the AMBER convention specifies, kcal/mol/Å, converted from eV/Å, as the velocities already were. Quantities the convention does not define keep GPUMD units, and the `units` attribute of each variable is specifying the unit.

The quantity parsing is shared with `dump_xyz` through `parse_utilities`, so the two keyword lists stay in step.
As a side effect a repeated quantity is now an error in `dump_xyz` too, as repeated `group` and `precision` already were.

`check_need_peratom_virial` also scans `dump_netcdf` lines now; without that, `dump_netcdf virial` would silently lose its PPPM reciprocal-space part.

Group selection moved from a CUDA kernel to the host, as in `dump_xyz`, so the grouped and whole-system cases share one packing path instead of needing a device buffer and a launch per quantity.

## Appending

An existing file is extended rather than replaced, matching `dump_xyz`, so rerunning GPUMD in the same directory continues the trajectory.
What decided between appending and creating was a static list of the names the process had written, which is empty at startup; that is replaced by a test of what is on disk, which covers both cases with one rule.

A NetCDF layout is fixed when the file is created, so a rerun whose atom count, precision, quantities, compression or group differ from the file stops instead of silently starting over.
The messages name the file, since removing or renaming it is what the user has to do next.
A file that does not record its quantities at all is reported the same way, at the comparison rather than at open, so that the more specific checks still speak first.

Note that `time` restarts at the beginning of each execution while `frame` keeps counting, so times in a file appended to across executions are not monotonic.

## Consolidation with dump_xyz

The two keywords have very similar purpose but differed in argument grammar, quantities supported, and the behavior on a rerun.
They now share all three.

The shared code lives in `parse_utilities`, which already held `parse_group` and `parse_precision` and now also holds the `DumpQuantities` struct and `parse_dump_quantity`.
Both keywords call all three.
A quantity added there appears in both at once, and neither can drift into accepting a name the other rejects. `dump_netcdf` also took over the append behavior and the host-side group selection that `dump_xyz` already had, so the grouped and whole-system cases go through one path.

For the user this means the two are close to interchangeable: apart from the file name, the same line works with either keyword, and the same trajectory can be written in text or binary form by changing one token.

One consequence for `dump_xyz`: a repeated quantity is now an error there too, as repeated `group` and `precision` already were.

Not shared, deliberately: each keyword keeps its own option loop, since `dump_xyz` has the trailing-star per-frame file name and `dump_netcdf` has `compression`; and the writing itself stays separate, one being self-describing text per frame and the other a fixed binary layout declared when the file is created.

## Notes

Tests follow the `dump_xyz` patterns, including a rejection table that asserts message text.
Every guard against a silent failure was confirmed by removing it and checking the matching test fails.

`dump_netcdf.cu` was not clang-format clean, so the mandated reformat reflows some lines unrelated to this change.

### DIFF
--- a/doc/gpumd/input_parameters/dump_netcdf.rst
+++ b/doc/gpumd/input_parameters/dump_netcdf.rst
@@ -53,7 +53,7 @@ Currently, the following optional arguments are accepted:
      * - :attr:`force`
        - ``forces``
        - ``(frame, atom, spatial)``
-       - eV/Å
+       - kcal/mol/Å
      * - :attr:`unwrapped_position`
        - ``unwrapped_coordinates``
        - ``(frame, atom, spatial)``
@@ -130,9 +130,12 @@ Requirements and specifications
   so that periodic geometry is preserved. Positions, velocities, forces and unwrapped positions
   are rotated as vectors, and the per-atom virial and the :term:`BEC` as rank-2 tensors, so that
   every quantity in a frame refers to the same axes.
-* The units of the ``forces``, ``potential_energy``, ``charge``, ``bec``, ``virial`` and ``mass``
-  variables are those GPUMD works in, and are given by the ``units`` attribute of each variable.
-  They are not the units the AMBER conventions specify for the variables AMBER defines.
+* The variables the AMBER conventions define, namely ``time``, ``coordinates``, ``velocities``,
+  ``forces``, ``cell_lengths`` and ``cell_angles``, carry the units those conventions specify, so
+  the velocities are written in Å/ps and the forces in kcal/mol/Å rather than in the units GPUMD
+  works in. This is what lets readers that trust the ``Conventions`` attribute, such as MDAnalysis,
+  open the file. The remaining variables are GPUMD extensions that the conventions do not cover and
+  carry GPUMD units. The ``units`` attribute of each variable is authoritative in either case.
 * For qNEP models the charge values are predicted by the model. For other models they are those
   specified in :ref:`model.xyz <model_xyz>` via ``charge:R:1``.
 * The Born effective charge (:term:`BEC`) requires a qNEP model trained with target :term:`BEC`,
--- a/doc/gpumd/input_parameters/dump_netcdf.rst
+++ b/doc/gpumd/input_parameters/dump_netcdf.rst
@@ -14,21 +14,30 @@ Syntax
 
 This keyword has the following format::
 
-  dump_netcdf <grouping_method> <group_id> <interval> <has_velocity> <filename> [{optional_args}]
+  dump_netcdf <interval> <filename> [{optional_args}]
 
-The :attr:`grouping_method` parameter selects one of the grouping methods defined by
-``group:I:number_of_grouping_methods`` in the :ref:`simulation model file <model_xyz>`.
-Grouping methods are numbered from 0. When :attr:`grouping_method` is non-negative,
-:attr:`group_id` selects a group label within that grouping method, and only the atoms
-belonging to that group are written. Both values must identify an existing, non-empty
-group. When :attr:`grouping_method` is negative, :attr:`group_id` is ignored and all atoms
-in the system are written.
-The :attr:`interval` parameter is the output interval (number of steps) of the atom positions. :attr:`has_velocity` can be 1 or 0, which means the velocities will or will not be included in the output. :attr:`filename` is the relative or absolute path of the output file.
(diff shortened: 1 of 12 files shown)


doc/gpumd/input_parameters/dump_netcdf.rst
```diff
@@ -92,7 +166,16 @@ Group output with compression
...
 
-  dump_netcdf 1 0 15 1 group.nc compression deflate 1
+  dump_netcdf 15 group.nc group 1 0 velocity compression deflate 1
+
+before the :ref:`run command <kw_run>`.
+
+Forces and per-atom virials
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To dump forces, per-atom potential energies and per-atom virials every 100 steps, one can add::
```
erhart1:
```suggestion
To dump forces, per-atom potential energies and per-atom virials every 100 steps with lossless deflate compression, one can add::
```

Kat543:
Works nicely. I have just a small comment that, when including the forces, the trajectory can't be read with MDAnalysis.Universe (`NotImplementedError: NETCDFReader currently assumes that the trajectory was written in units of eV/angstrom instead of kilocalorie/mole/angstrom`)

But reading the trajectory with netCDF4.Dataset works fine.

erhart1:
> Works nicely. I have just a small comment that, when including the forces, the trajectory can't be read with MDAnalysis.Universe (`NotImplementedError: NETCDFReader currently assumes that the trajectory was written in units of eV/angstrom instead of kilocalorie/mole/angstrom`)
> 
> But reading the trajectory with netCDF4.Dataset works fine.

The forces were not converted to AMBER units before writing. This is fixed now.

erhart1:
@brucefan1983 this is ready for review now

brucefan1983:
PR #1631 introduced GPU-side gathering for grouped NetCDF output, so only the selected group was copied back to the host. This PR appears to replace that with full-system D2H copies followed by host-side selection. For a small group in a large system this can substantially increase the transfer cost.

erhart1:
> PR #1631 introduced GPU-side gathering for grouped NetCDF output, so only the selected group was copied back to the host. This PR appears to replace that with full-system D2H copies followed by host-side selection. For a small group in a large system this can substantially increase the transfer cost.

You are completely right. I had meant to still fix that and forgot. It is done now in [ee40314](https://github.com/brucefan1983/GPUMD/pull/1648/commits/ee4031479b045713ae5852895015e6d6fb4d95c8).

PR #1661, opened 2026-08-22
Write the dump_netcdf mass every frame

erhart1:
## Summary

`dump_netcdf` defined the `mass` variable over the `atom` dimension alone and wrote it once, when the file was created.
A Monte Carlo trial changes the species of a site and with it the mass, so from the first accepted trial the mass in the file no longer matched the `type` variable, which already carries a frame dimension.
Any analysis that takes the mass from the trajectory file, for instance a kinetic energy or a mass-weighted velocity autocorrelation, was wrong for runs that use the `mc` keyword.

`mass` now carries a frame dimension and is written every frame, staged from the device array `atom.mass` through the same path as the other per-atom quantities, so a group selection gathers only the atoms it writes and the variable picks up chunking and deflate when compression is requested.
`dump_xyz` already behaves this way.

## Breaking changes

The `mass` variable is `(frame, atom)` instead of `(atom)`, so a reader that indexes it by atom alone needs to select a frame.

Appending to a trajectory whose `mass` has no frame dimension is refused with a message naming the file, rather than failing inside netcdf-c.
Such a file needs to be renamed or removed to start a new trajectory.

## Testing

Three tests were added to `tests_pytest/test_io_dump_commands.py`:

* the mass follows the species of a site across `mc canonical` trials, with the mass and the type of every site agreeing in every frame and the canonical composition conserved;
* with a group selected, the masses in the file are exactly those of the atoms of that group, checked against a whole-system dump of the same run;
* appending to a file whose `mass` is indexed by atom alone is refused.

The `mass` entry of the existing quantity-layout table was updated to `(frame, atom)`.

`tests_pytest` passes in full on my workstation (RTX 3080Ti) with a NetCDF-enabled build: 452 passed and 145 skipped in `test_io_dump_commands.py`, and 294 passed and 78 skipped across the rest of the fast suite.
`dump_netcdf.cu` compiles without warnings under the project warning flags.

### DIFF
--- a/doc/gpumd/input_parameters/dump_netcdf.rst
+++ b/doc/gpumd/input_parameters/dump_netcdf.rst
@@ -76,15 +76,17 @@ Currently, the following optional arguments are accepted:
        - eV
      * - :attr:`mass`
        - ``mass``
-       - ``(atom)``
+       - ``(frame, atom)``
        - amu
      * - :attr:`group_labels`
        - ``group_labels``
        - ``(atom, grouping_method)``
        - --
 
-  The mass and the group labels cannot change during a run and are therefore written once,
-  without a frame dimension.
+  The group labels are fixed by the :ref:`simulation model file <model_xyz>` and are therefore
+  written once, without a frame dimension.
+  The mass is written every frame, because a :term:`MC` trial performed with the
+  :ref:`mc keyword <kw_mc>` changes the species of a site and with it the mass.
   The ``group_labels`` variable holds one label per grouping method, giving the group each atom
   belongs to.
   The rank-2 tensors are stored row-major, so ``virial[frame, atom, i, j]`` is the *ij*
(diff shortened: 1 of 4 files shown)


doc/gpumd/input_parameters/dump_netcdf.rst
```diff
@@ -76,15 +76,17 @@ Currently, the following optional arguments are accepted:
...
        - ``group_labels``
        - ``(atom, grouping_method)``
        - --
 
-  The mass and the group labels cannot change during a run and are therefore written once,
-  without a frame dimension.
+  The group labels are fixed by the :ref:`simulation model file <model_xyz>` and are therefore
+  written once, without a frame dimension.
+  The mass is written every frame, because a :term:`MC` trial performed with the
+  :ref:`mc keyword <kw_mc>` changes the species of a site and with it the mass.
```
erhart1, line 89:
```suggestion
  :ref:`mc keyword <kw_mc>` changes the species of a site and with it the mass.
  Note that this also means that a lot of data is duplicated.
  Since the mass can usually be attributed based on the atom type,
  in most cases it is recommended *not* to write the mass but to rather assign it
  after the fact based on the atom type in order to save storage space.
```

brucefan1983:
Nice work. I think it is done?

